PR #2, opened 2020-07-17
More info about examples

danielballan:


danielballan:
Only changes are to docs. Self-merging.

PR #4, opened 2020-07-29
Avoid terminate

### LINKED ISSUE #3: Revisit how to stop the DataLoader thread
Two things about this, per @ronpandolfi:

https://github.com/bluesky/bluesky-widgets/blob/38b7f1aa6c26a4b9e59a71acfdc446a377755bca/bluesky_widgets/qt/search_results.py#L91

* Don't use `terminate`. Take the approach (which we use a lot in caproto) of setting an `Event` that we check on periodically. Add a timeout for the `get()`.
* Connect setting this `Event` to both `destroyed` _and_ `QApplication.aboutToQuit`.

danielballan:
Closes #3 

Debug logging shows that the DataLoader does shutdown cleanly as intended when
the app is closed. Pending further comment from @ronpandolfi on #3 about
whether we need to reach up to an application signal as well to cover some
scenario we haven't thought of.

### DIFF
--- a/bluesky_widgets/qt/search_results.py
+++ b/bluesky_widgets/qt/search_results.py
@@ -34,7 +34,16 @@ class DataLoader(QThread):
     """
 
     def __init__(self, queue, get_data, data, data_changed, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+        # Because the run method contains an infinite loop (standing by to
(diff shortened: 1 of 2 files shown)


danielballan:
Power-cycling CI to get CI fixes from #5

PR #5, opened 2020-07-29
Allow pyside2 CI failures for now.

danielballan:
@dmgav created comprehensive CI matrix in #1. The pyside2 builds have not yet
ever passed, and we have been ignoring those failures in the short term.

This PR formalizes that, so that the feedback from the GitHub UI is useful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,12 +45,6 @@ jobs:
       env: QT="pyside2"
     - python: 3.8
       env: QT="pyside2"
-    - os: osx
-      language: generic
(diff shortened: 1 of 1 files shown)


dmgav:
Tests with PySide2 with OSX seem to pass, so may be we can leave them as required.

danielballan:
Good point, @dmgav. Done.

I also observed an unexpected error while installing numpy on one of the PyQt5 buidls. We'll see if 3b42330 fixes it.

PR #10, opened 2020-08-04
Refactor DataLoader again; remove usage of QThread entirely.

### LINKED ISSUE #7: Refactor to use QThreadPool and QRunnable
This is now the preferred when it fits, offered a more constrained but simpler API. https://www.learnpyqt.com/courses/concurrent-execution/multithreading-pyqt-applications-qthreadpool/

Napari has some interesting tooling for making threadpool workers out of functions and generators, and a nice API for waiting on cleanup at shutdown (up to some time limit) and raising a more useful error if the workers refuse to die. https://github.com/napari/napari/blob/master/napari/_qt/threading.py

### LINKED ISSUE #9: Confirm that QThreads are actually closing themselves
In response to #4 , I'm not confident that connecting a signal directly to requestInterruption works without an event loop in the worker thread. We could either confirm this is working or instead bind to a method on the parent widget to call `thread.requestInterruption()`

There is some good discussion of the tricks inherent to QThreads here:
https://wiki.qt.io/Threads_Events_QObjects

(See "DOs and DON'Ts" and Per-thread event loop)

Also fun:
https://woboq.com/blog/qthread-you-were-not-doing-so-wrong.html

danielballan:
Issues addressed:

- Modern best practice for Qt is to avoid using `QThread` directly unless you really need the lower-level methods that they expose. Instead, `QThreadPool` and `QRunnable` are [recommended](https://www.learnpyqt.com/courses/concurrent-execution/multithreading-pyqt-applications-qthreadpool/), and they are exposed nicely in Python.
- As @HarinarayanKrishnan kindly pointed out, passing the `dataChanged` signal from `_SearchResultsModel` into the `__init__` of `DataLoader` was bad practice. better to expose a signal out from `DataLoader` and connect it up to `dataChanged`.
- We also don't need to be sharing a Python `queue.Queue` between `_SearchResultsModel` and `DataLoader`. The `DataLoader` can just expose a `request` method for accepting new work and manage its queue (or whatever) internally

Implementation:

Incidentally, napari follows the recommendation to use `QThreadPool` and `QRunnable` and has a uesful module of mid-level `Worker` abstractions similar to the ones in the linked article. They do things like "Run this function on a thread," or "Run this generator on a thread with two-way communication," and they provide a nice cleanup API. I have vendored that module into bluesky-widgets.

All of the cleverness around `parent` and `weakref` that entered into #4 is no longer needed.

We connect the `wait_for_workers_to_quit` function up to the global `QApplication.aboutToQuit` signal --- not far off from @ronpandolfi initial impulse, IIRC. (Since the workers in question are running on a global `QThreadPool`, our earlier qualms about reaching way up to an application-wide signal to kill a specific `QThread` no longer apply.)

Runs (and exits cleanly) in our test suite, our examples, and in Xi-cam, with the addition of a line or two to connect the cleanup to `aboutToQuit`.

Closes #7 
Closes #9


bluesky_widgets/qt/search_results.py
```diff
@@ -34,20 +40,22 @@ class DataLoader(QThread):
       viewer from the data cache.
     """
 
-    def __init__(self, queue, get_data, data, data_changed, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self._queue = queue
+    def __init__(self, get_data, data, *args, **kwargs):
+        super().__init__(*args, SignalsClass=DataLoaderSignals, **kwargs)
         self._get_data = get_data
         self._data = data
```
HarinarayanKrishnan, line 46: Looks like self._data is being shared across two threads (DataLoader) and GuiThread.  Are there guarantees that both threads won't touch data simultaneously? (Maybe I missed something, just confirming to be sure.)

bluesky_widgets/qt/search_results.py
```diff
@@ -34,20 +40,22 @@ class DataLoader(QThread):
...
-        self._data_changed = data_changed
+        self._queue = queue.Queue()
 
-    def run(self):
+    def request(self, work):
```
HarinarayanKrishnan, line 49: Can def request be turned into a slot, that way you can emit from _fetchMore a signal that is received by the request slot? Otherwise you are calling into request in DataLoader which touches self._queue from another thread

bluesky_widgets/qt/search_results.py
```diff
@@ -136,7 +123,7 @@ def _fetch_data(self, index):
             return self._data[index]
         else:
             self._data[index] = LOADING_PLACEHOLDER
-            self._request_queue.put(index)
+            self._data_loader.request(index)
```
HarinarayanKrishnan, line 126: I don't think _fetch_data is necessary. You could instead in data() in line 174/187, directly emit a list of QModelIndex (or QModelIndexList) to DataLoader.request(). Then when data arrives the DataLoader will then do what it is doing now which is emitting back to self.dataChanged index and value. Also, removing this function will remove self._data being used in two places and keep the two objects separated.

bluesky_widgets/qt/threading.py
```diff
@@ -0,0 +1,625 @@
...
+# In the future, this could wrap any API, or a pure python threadpool.
+
+
+def set_max_thread_count(num: int):
+    """Set the maximum number of threads used by the thread pool.
+
+    Note: The thread pool will always use at least 1 thread, even if
+    maxThreadCount limit is zero or negative.
+    """
+    QThreadPool.globalInstance().setMaxThreadCount(num)
```
HarinarayanKrishnan, line 351: No issue here, Just a high level inquiry/concern: As a stand alone application exposing access to setting these variables is fine, but when it is underneath another larger application. Could this potential for issues? For example: Napari sets set_max_thread_count() then some developer sees this in bluesky_widgets and sets the same thing using a different function in a different location. Now there is a conflict. Not saying this will happen, also I am not entirely sure what a good solution looks like. But potentially exposing functions that has side-effects from dependencies will mess with the overall program execution. At least, using a global shared pool definitely has some pros versus everyone creating their own local threadpool and producing crazy overhead in thread switching.

bluesky_widgets/qt/search_results.py
```diff
@@ -69,7 +77,7 @@ def run(self):
             # requires some measurement. For now, the performance of this
             # implementation is acceptable. Most important, the application
             # does not lock up at all during data loading.
-            self._data_changed.emit(index, index, [])
+            self._signals.data_changed.emit(index, index, [])
```
HarinarayanKrishnan, line 80: It might make sense to not perform self._data[index] = item and rather emit the item with the index and perform the assignment back in the main thread

bluesky_widgets/qt/search_results.py
HarinarayanKrishnan: I realized that you are running a QRunnable object and QThread anymore, which complicates this idea slightly. The recommendation from Qt in this case is that you are on your own to guard your data using Mutexes.

bluesky_widgets/qt/search_results.py
danielballan: 1. The GUI thread mutates `_data` by adding a placeholder string (`"..."`).
2. The GUI thread pushes work to the DataLoader thread.
3. Sometime later, the DataLoader thread mutates `_data` again, replacing the placeholder with the actual value.

So far, so good! However, your comment has made me realize that this tidy story breaks down when the catalog is reloaded. At that point `_data` is cleared on the main thread, but there might still be work in the DataLoader's queue. In that scenario, stale values could  get filled in and the true values would never be loaded. This is a critical problem.

bluesky_widgets/qt/search_results.py
danielballan: I don't fully understand how using the QThreadPool instead of QThread complicates or changes things. Can you explain more? In any case, I'm sold that using signals/slots would be better and is worth doing anyway.

bluesky_widgets/qt/search_results.py
danielballan: I'm sold. Thanks for pointing this out.

bluesky_widgets/qt/threading.py
danielballan: It's a good point. Ultimately getting this right will be up to the person who is managing the QApplication instance. I think the best we can do is leave clear code comments in examples where we use this.

danielballan:
On a phone call, @HarinarayanKrishnan pointed out that this PR was mixing two different modalities. A threadpool, particularly a global shared threadpool, is intended for singleshot tasks, not a long-lived infinitely-looping worker which would hog a spot in the pool forever. It would be better to choose between:

1. A long running `QThread` worker with an event loop that accepts new work from a signal/slot
2. Creating short-lived `QRunnable` workers in the global `QThreadPool` that process fixes batches of work specified when they are created and then stop when that work is complete

I prefer (2) because it provides a clear solution for what to do when the model refreshes: you disconnect the workers, cancel them (just to avoid unncessary work), and create new workers as needed. I also prefer (2) because I like using the more constrained abstraction of `QRunnable` if we can make it fit well.

The implementation came together easily enough, and it removed yet more complexity so it feels like a step in the right direction.

Performance is satisfactory and shutdown is clean, both in the stand-alone example and when embedded in Xi-CAM. We can tune the batch sizes and processing intervals to improve performance further.

bluesky_widgets/qt/threading.py
```diff
@@ -0,0 +1,625 @@
...
+        self._pause_interval = 0.01
+
+    def work(self) -> None:
+        """Core event loop that calls the original function.
+
+        Enters a continual loop, yielding and returning from the original
+        function.  Checks for various events (quit, pause, resume, etc...).
+        (To clarify: we are creating a rudimentary event loop here because
+        there IS NO Qt event loop running in the other thread to hook into)
+        """
```
HarinarayanKrishnan, line 277: I am not entirely sure what the main reason for this comment. Maybe I am not understanding what it means when the comment say there is NO Qt event loop in the other thread. You can always construct things in a way you can create a QEventLoop at the appropriate level to hook it.

bluesky_widgets/qt/search_results.py
```diff
@@ -94,55 +53,55 @@ def __init__(self, model, *args, **kwargs):
...
+        self._active_workers = set()
 
-        # The DataLoader thread is not started here. It will be started if/when
-        # we need it for the first time.
-        self._data_loader = data_loader
+        # Start a timer that will periodically load any data queued up to be loaded.
+        LATENCY = 100  # ms
+        self._data_loading_timer = QTimer(self)
+        self._data_loading_timer.timeout.connect(self._process_work_queue)
+        self._data_loading_timer.start(LATENCY)
```
HarinarayanKrishnan, line 67: This is great. Entirely optional alternative strategy, is to create a single shot Timer and in self._process_work_queue (restart the timer). That way you don't run into an issue where Timer keeps firing _process_work_queue one after another in case the _process_work_queue ever takes greater than LATENCY

danielballan: Good point. @tacaswell pointed me in this direction in some much earlier draft of this, but it fell out of my brain. Thanks for the reminder.

danielballan: Addressed in 6ea3841

PR #13, opened 2020-08-06
Jr fork/ergo serch input

JulReinhardt:
simple implementation to add extra time selector buttons for more ergonomic selection of catalogs

### DIFF
--- a/bluesky_widgets/qt/search_input.py
+++ b/bluesky_widgets/qt/search_input.py
@@ -1,5 +1,5 @@
 import time
-from datetime import datetime, timedelta
+from datetime import timedelta
 from qtpy.QtCore import QDateTime
 from qtpy.QtWidgets import (
     QButtonGroup,
@@ -9,7 +9,6 @@ from qtpy.QtWidgets import (
     QFormLayout,
     QRadioButton,
     QGridLayout,
-    QButtonGroup,
 )
 
 
@@ -75,7 +74,7 @@ class QtSearchInput(QWidget):
         # Changes to the GUI update the model.
         self.since_widget.dateTimeChanged.connect(self.on_since_view_changed)
         self.until_widget.dateTimeChanged.connect(self.on_until_view_changed)
-        #TODO: check what refresh button is doing
+        # TODO: check what refresh button is doing
         self.refresh_button.clicked.connect(self.model.events.reload)
         self.refresh_button.clicked.connect(self.on_refresh_clicked)
         # Changes to the model update the GUI.
@@ -103,15 +102,18 @@ class QtSearchInput(QWidget):
         now = time.time()
         if isinstance(event.date, timedelta):
             if event.date == timedelta(days=1):
-                self.since_widget.setDateTime(QDateTime.fromSecsSinceEpoch(now + timedelta(days=-1).total_seconds()))
+                self.since_widget.setDateTime(
+                    QDateTime.fromSecsSinceEpoch(now + timedelta(days=-1).total_seconds()))
                 self.until_widget.setDateTime(QDateTime.fromSecsSinceEpoch(now))
                 self.today_widget.setChecked(True)
             if event.date == timedelta(days=30):
-                self.since_widget.setDateTime(QDateTime.fromSecsSinceEpoch(now + timedelta(days=-30).total_seconds()))
+                self.since_widget.setDateTime(
+                    QDateTime.fromSecsSinceEpoch(now + timedelta(days=-30).total_seconds()))
                 self.until_widget.setDateTime(QDateTime.fromSecsSinceEpoch(now))
                 self.days_widget.setChecked(True)
             if event.date == timedelta(minutes=60):
-                self.since_widget.setDateTime(QDateTime.fromSecsSinceEpoch(now + timedelta(minutes=-60).total_seconds()))
+                self.since_widget.setDateTime(
+                    QDateTime.fromSecsSinceEpoch(now + timedelta(minutes=-60).total_seconds()))
                 self.until_widget.setDateTime(QDateTime.fromSecsSinceEpoch(now))
                 self.hour_widget.setChecked(True)
         else:
@@ -142,7 +144,6 @@ class QtSearchInput(QWidget):
         # TODO: since/until widget should update immediately when clicking refresh
         self.uncheck_radiobuttons()
         # TODO: check which RButton is selected to update that range
-
 
     def on_select_today(self):
         self.model.since = timedelta(days=1)
--- a/bluesky_widgets/qt/search_input.py
+++ b/bluesky_widgets/qt/search_input.py
@@ -86,6 +86,7 @@ class QtSearchInput(QWidget):
         self.since_widget.dateTimeChanged.connect(self.on_since_view_changed)
         self.until_widget.dateTimeChanged.connect(self.on_until_view_changed)
         self.refresh_button.clicked.connect(self.model.events.reload)
+        self.refresh_button.clicked.connect(self.on_refresh_clicked)
         # Changes to the model update the GUI.
         self.model.events.since.connect(self.on_since_model_changed)
         self.model.events.until.connect(self.on_until_model_changed)
@@ -123,7 +124,10 @@ class QtSearchInput(QWidget):
         qdatetime.setSecsSinceEpoch(event.date)
         self.until_widget.setDateTime(qdatetime)
 
-
+    def on_refresh_clicked(self):
+        self.now = time.time()
+        #TODO: since/until widget should update immediately when clicking refresh
+        # check which RButton is selected to update that range
 
     def set_timerange(self, timerange):
         self.since_widget.setDateTime(QDateTime.fromSecsSinceEpoch(self.now - timerange))
--- a/bluesky_widgets/qt/search_input.py
+++ b/bluesky_widgets/qt/search_input.py
@@ -51,32 +51,32 @@ class QtSearchInput(QWidget):
         default_period_layout.addWidget(self.hour_widget, 2, 1, 1, 2)
         self.layout().addRow("When:", default_period_layout)
 
-        # Restrict acceptable timedelta values
-        self.allowed = {timedelta(days=-1), timedelta(days=-30), timedelta(minutes=-60), timedelta(days=-7),
-                        timedelta(days=-365)}
-        def time_validator(since=None, until=None):
-            """
-            Enforce that since and until are values that a UI can represent.
-            This is an example similar to what will be used in the Qt UI.
-            """
-            now = timedelta()
-            if isinstance(since, timedelta):
-                if not (until is None or until == now):
-                    raise ValueError(
-                        "This UI cannot express since=timedelta(...) unless until "
-                        "is timedelta() or None."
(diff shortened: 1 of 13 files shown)


danielballan:
I merged #14 this morning, which introduced many conflicting changes with this PR. To make up for that, I have done the rebase for you and force-pushed. I hope I reconciled the changes correctly; please check my work. :-)

Will comment later on your question on Slack about refreshing. Thinking....

bluesky_widgets/qt/search_input.py
```diff
@@ -16,54 +26,138 @@ def __init__(self, model, *args, **kwargs):
...
+        self.refresh_button.clicked.connect(self.on_refresh_clicked)
         # Changes to the model update the GUI.
         self.model.events.since.connect(self.on_since_model_changed)
         self.model.events.until.connect(self.on_until_model_changed)
 
+        # connect QRadioButtons and change date dropdowns (since/until widgets) accordingly
+        self.today_widget.clicked.connect(self.on_select_today)
+        self.hour_widget.clicked.connect(self.on_select_lasthour)
+        self.days_widget.clicked.connect(self.on_select_30days)
+        self.all_widget.clicked.connect(self.on_select_all)
```
danielballan, line 86: Here in `__init__` we should configure `model` to only accept certain timedelta values (the ones which the particular UI can display) and reject any others with a `ValueError`. I sketched the code to do so in a test: I think all that's left to do is copy/paste [this test code](https://github.com/bluesky/bluesky-widgets/pull/13/files#diff-b58e3b3ad1d5c95fdfb98a50d43a9972R88-R113) here in `__init__`.

bluesky_widgets/qt/search_input.py
```diff
@@ -16,54 +26,138 @@ def __init__(self, model, *args, **kwargs):
...
-        qdatetime.setSecsSinceEpoch(self.model.until)
+        qdatetime.setSecsSinceEpoch(self.model.until.timestamp())
         self.until_widget.setDateTime(qdatetime)
 
         # Changes to the GUI update the model.
         self.since_widget.dateTimeChanged.connect(self.on_since_view_changed)
         self.until_widget.dateTimeChanged.connect(self.on_until_view_changed)
-        self.refresh_button.clicked.connect(self.model.events.reload)
+        # TODO: check what refresh button is doing
+        # self.refresh_button.clicked.connect(self.model.events.reload)
```
danielballan, line 76: Hm, I agree that this is subtle. After thinking about this for bit, I believe the correct "reactive" approach is to connect the button directly to `self.model.request_reload`. All the button should do is tell the model to reload.

Then, create a new method `on_reload` on this class and, in `__init__`, connect it to both `self.model.events.reload` (which fires when we have refreshed content in the same catalog) and `self.model.events.query` (which fires when we have a new catalog). Inside `on_reload`, inspect `self.model.since` and `self.model.until`. If they are timedeltas, it should compute the absolute time relative to "now" and update the since/until UIs accordingly, similar to the code in `on_since_model_changed`. In this case, we won't have `event.date` passed in, but we can consult the model directly by looking at `self.model.since`. The value of `since` will _not_ have changed, but we will reevaluate its meaning relative to "now".

bluesky_widgets/qt/search_input.py
```diff
@@ -16,54 +26,138 @@ def __init__(self, model, *args, **kwargs):
...
         self.model.events.since.connect(self.on_since_model_changed)
         self.model.events.until.connect(self.on_until_model_changed)
 
+        # connect QRadioButtons and change date dropdowns (since/until widgets) accordingly
+        self.today_widget.clicked.connect(self.on_select_today)
+        self.hour_widget.clicked.connect(self.on_select_lasthour)
+        self.days_widget.clicked.connect(self.on_select_30days)
+        self.all_widget.clicked.connect(self.on_select_all)
+
+    def on_refresh_clicked(self):
```
danielballan, line 88: So, per my long comment above, this method can just go away.

bluesky_widgets/qt/search_input.py
```diff
@@ -16,54 +26,138 @@ def __init__(self, model, *args, **kwargs):
...
+    def on_select_lasthour(self):
+        self.model.since = timedelta(minutes=60)
+        self.model.until = timedelta()
+
+    def on_select_30days(self):
+        self.model.since = timedelta(days=30)
+        self.model.until = timedelta()
+
+    def on_select_all(self):
+        # self.model.until = None
```
danielballan, line 150: This should update `self.model` but not touch the UI directly, just like the other `on_select_*` methods. I'm fine with using `0` as the default for now, though I also like the idea of a cheeky choice like Galileo's birthday or something. We can always iterate later. :-D

JulReinhardt: done, see commit

bluesky_widgets/qt/search_input.py
JulReinhardt: yes, done.

bluesky_widgets/qt/search_input.py
JulReinhardt: > Then, create a new method `on_reload` on this class and, in `__init__`, connect it to both `self.model.events.reload` (which fires when we have refreshed content in the same catalog) and `self.model.events.query` (which fires when we have a new catalog). 

What do you mean by 'on this class'? The SearchInput class?

danielballan: Right, exactly as in the commit you just added. 👍

JulReinhardt: Ok, I don't get the connection right yet, as model.events.reload and query don't lead to the same event object as model.events.since or until. I get the error: AttributeError: 'Event' object has no attribute 'date'

bluesky_widgets/qt/search_input.py
JulReinhardt: Ok, this restriction is only needed for input from the CL. Is the simple copy/paste sufficient for your purpose?

JulReinhardt: The 'all' timedelta does not fit too well here, since all is an arbitrary definition we (or I atm made, did you figure out what date I chose? ;-))

JulReinhardt:
After my last cleanup commit, now looking at the timedelta validator: I'm not sure why it is needed here, because the RButtons in the gui would only be checked, when one of the 4 timedeltas defined for these buttons are set. So if someone tried to set a different timedelta via CL, it would simply changes since and until widget but not check a RadioButton, so what's the matter for the validator then?

danielballan:
If someone set a custom timedelta via the CL, the UI would be in a state where, looking at it, you wouldn't be able to tell whether the since/until were absolute or relative. But maybe that's not such a big problem; maybe it's OK to allow some things from the CL that aren't possible to do through the UI and assume that if someone did this they would understand why since/until would update themselves whenever the catalog refreshed. It's a good point.

I would be OK with omitting the validator, at least until we hear from some user who got confused by this "secret feature". :-D

danielballan:
P.S. Thanks for fixing he signs in my tests, @JulReinhardt. :-D

JulReinhardt:
Yes, true the UI would not determine between relative and absolute, but why does it matter as long as the since/until display are correctly displayed?

bluesky_widgets/qt/search_input.py
```diff
@@ -16,54 +26,168 @@ def __init__(self, model, *args, **kwargs):
...
 
+        # connect QRadioButtons and change date dropdowns (since/until widgets) accordingly
+        self.today_widget.clicked.connect(self.on_select_today)
+        self.hour_widget.clicked.connect(self.on_select_lasthour)
+        self.days_widget.clicked.connect(self.on_select_30days)
+        self.all_widget.clicked.connect(self.on_select_all)
+
+    def on_reload(self, event):
+        now = time.time()
+        if isinstance(event.date, timedelta):
```
danielballan, line 118: As you know, this is connected to the `self.model.events.reload` signal and the `self.model.events.query` signal. Those signals event `Event`s that do not have a date attached. (They are effectively just a timing signal: "It happened now!" with no extra content.) So, instead of `event.date` we need to use `self.model.since` here.

bluesky_widgets/qt/search_input.py
danielballan: Hmm...is that 1866?

bluesky_widgets/qt/search_input.py
danielballan: Oh, I understand now. See https://github.com/bluesky/bluesky-widgets/pull/13#discussion_r477485246

bluesky_widgets/qt/search_input.py
JulReinhardt: 1815-12-10... A birthday

bluesky_widgets/qt/search_input.py
```diff
@@ -113,14 +113,14 @@ def time_validator(since=None, until=None):
         self.days_widget.clicked.connect(self.on_select_30days)
         self.all_widget.clicked.connect(self.on_select_all)
 
-    def on_reload(self, event):
+    def on_reload(self):
```
JulReinhardt, line 116: What input argument is needed here?

danielballan: The model will call it when a generic `Event` (no `date` in this case, but some general info about where it came from). I think it's passed as a positional argument so `on_reload(self, literally_anything)` should work  but `on_reload(self, event)` would be most readable.

So --- we have to accept `event` as an argument because of how signaling works, but in this case we aren't going to _do_ anything with it.

bluesky_widgets/qt/search_input.py
danielballan: Oh I see that's relative to 1970 of course, not relative to today. Ada!

bluesky_widgets/qt/search_input.py
JulReinhardt: Ok... I will accept this :)

JulReinhardt:
My suggestions for now are committed.
I commented the timedelta() restrictions out for now, since I don't see a clash. But happy to have it back in.

danielballan:
I think commented-out is a good state to leave them in. If it becomes a problem, we'll have a solution already thought-out and ready. If it's never a problem, we can delete the comment someday.

JulReinhardt:
Ok, Travis Check failed (on mac)? So are we ready for merge?

danielballan:
The `flake8` build failed too. Mind patching these little nit-picks from `flake8` up so it's happy?

```
./bluesky_widgets/qt/search_input.py:13:1: F401 '..components.search.search_input.SearchInput' imported but unused
./bluesky_widgets/qt/search_input.py:15:1: E302 expected 2 blank lines, found 1
./bluesky_widgets/qt/search_input.py:182:5: E303 too many blank lines (3)
```

danielballan:
The Mac builds fail while trying to install `numcodecs`, not related to this PR, so we can address that separately (now tracked in #19) and proceed with merging this once `flake8` is mollified. :rocket:

danielballan:
The board is green minus the known unrelated OSX build failures noted in #19. Thanks @JulReinhardt!

danielballan:
image

PR #14, opened 2020-08-07
Fixes to the databroker browser layout design

ronpandolfi:
- Better disowning of widgets
- Tighter and simpler layout


danielballan:
Before

After

bluesky_widgets/qt/searches.py
```diff
@@ -28,13 +28,13 @@ def __init__(self, model, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.model = model
 
-        self.layout = QVBoxLayout()
```
danielballan, line 31: Yikes. I should have learned not to do this earlier! Thanks for catching it.

PR #16, opened 2020-08-11
Run tree

cryos:
This adds a simple tree view that looks something like 
image
when looking at a sample run. It still needs more work, but sharing early so that I can iterate on it.

### DIFF
--- a/bluesky_widgets/_qt/qt_main_window.py
+++ b/bluesky_widgets/_qt/qt_main_window.py
@@ -53,7 +53,8 @@ class Window:
         self._qt_center = QWidget(self._qt_window)
 
         self._qt_window.setCentralWidget(self._qt_center)
-        self._qt_window.setWindowTitle(self.qt_widget.model.title)
+        if hasattr(self.qt_widget.model, 'title'):
+            self._qt_window.setWindowTitle(self.qt_widget.model.title)
         self._qt_center.setLayout(QHBoxLayout())
         self._status_bar = QStatusBar()
         self._qt_window.setStatusBar(self._status_bar)
--- a/bluesky_widgets/examples/qt_run_tree_view.py
+++ b/bluesky_widgets/examples/qt_run_tree_view.py
@@ -1,14 +1,14 @@
 """
-Select some runs and click the button. Their IDs will be printed to the
-terminal. In a real application, this could kick off data processing, export,
-or visualization.
+Displays a BlueskyRun in a QTreeView lazily populating the view. In a real
+application this would be used in conjunction with a data browser/search box or
+on the command line to explore runs.
 """
 from bluesky_widgets.qt import Window
 from bluesky_widgets.qt import gui_qt
 from bluesky_widgets.qt.run_tree import QtTreeView
 from bluesky_widgets.examples.utils.generate_msgpack_data import get_catalog
 
-from qtpy.QtWidgets import QPushButton, QVBoxLayout, QWidget
+from qtpy.QtWidgets import QWidget
 
 
 class Views(QWidget):
@@ -19,9 +19,6 @@ class Views(QWidget):
     def __init__(self, *, show=True, title=""):
         super().__init__()
         self.title = title
-
-        # ...and one listing any and all catalogs discovered on the system.
-        from databroker import catalog
 
         # You can set your own catalog/run here or use this synthetic data.
         self._run = get_catalog()[-1]
@@ -42,6 +39,7 @@ def main():
     print(__doc__)
     with gui_qt("Example Application"):
         views = Views(title="Example Application")
+        views.show()
 
 
 if __name__ == "__main__":
--- a/bluesky_widgets/examples/qt_run_tree_view.py
+++ b/bluesky_widgets/examples/qt_run_tree_view.py
@@ -1,0 +1,48 @@
+"""
+Select some runs and click the button. Their IDs will be printed to the
+terminal. In a real application, this could kick off data processing, export,
+or visualization.
+"""
+from bluesky_widgets.qt import Window
+from bluesky_widgets.qt import gui_qt
+from bluesky_widgets.qt.run_tree import QtTreeView
+from bluesky_widgets.examples.utils.generate_msgpack_data import get_catalog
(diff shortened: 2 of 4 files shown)


bluesky_widgets/examples/qt_run_tree_view.py
```diff
@@ -0,0 +1,48 @@
+"""
+Select some runs and click the button. Their IDs will be printed to the
```
danielballan, line 2: Clearly this needs updating. Just marking it so we don't forget before merging.

bluesky_widgets/qt/run_tree.py
```diff
@@ -0,0 +1,195 @@
...
+    QModelIndex,
+    QTimer,
+    Qt,
+)
+from qtpy.QtWidgets import QAbstractItemView, QHeaderView, QTreeView
+
+from databroker import catalog
+from databroker.core import BlueskyEventStream
+
+class RunTree(object):
```
danielballan, line 13: Subclassing from `object` was necessary in Python 2 to opt in to so-called "new style" classes, but in Python 3-only codebases it's no longer necessary. It's a no-op.

bluesky_widgets/qt/run_tree.py
```diff
@@ -0,0 +1,195 @@
...
+    """Lazily populate the tree as data is requested. """
+    def __init__(self, run):
```
danielballan, line 15: A long-term comment, not implying the need for any changes here (for the present):

I am intrigued by applying the pattern `QtThing(model)` everywhere in this library, as we started to do with `QtSearch(seach_model: SearchModel)` and so on with its sub-components. In this case, I think the `model` would signal things like "just received a new {stream, descriptor, event, stop document}" if the `BlueskyRun` instance in question were a live one, and the tree would update itself accordingly.

It seems at least _plausible_ that the `model` in question here could just be a `BlueskyRun` with observable signals, provided by subclassing or by injection. I have some yet-to-be-finished-or-pushed code in this direction that I think will match up with this well.

However, if we want to be able to "hot swap" which Run we are viewing in this Tree, then maybe we want the extra layer of indirection of a `model` which has-a BlueskyRun, and can have different BlueskyRuns assigned to it during its lifecycle.

bluesky_widgets/qt/run_tree.py
```diff
@@ -0,0 +1,195 @@
...
+        self.run = run
+        self.children = []
+
+        uid = RunNode(self.run, 'uid', self.run.metadata['start']['uid'], None, self)
+        start = RunNode(self.run, 'start', 'dict', self.run.metadata['start'], self)
+        start.num_children = len(self.run.metadata['start'])
+        stop = RunNode(self.run, 'stop', 'dict', self.run.metadata['stop'], self)
+        stop.num_children = len(self.run.metadata['stop'])
+        streams = RunNode(self.run, 'streams', 'events (' + str(len(self.run)) + ')', None, self)
```
danielballan, line 24: I would nudge you toward embracing f-strings over `+` for this kind of construction because they are more readable and performant --- i.e. `f"events ({len(self.run)})"`.

Separately, I guess this is really number of _streams_ so the word "events" might be unhelpful here. Maybe we just want `f"({len(self.run)})"`.

bluesky_widgets/qt/run_tree.py
```diff
@@ -0,0 +1,195 @@
...
+                self.children.append(child)
+        elif self.data and isinstance(self.data, BlueskyEventStream):
+            # For now just display the keys in the stream.
+            node = RunNode(self.run, 'metadata', 'dict', self.data.metadata, self)
+            node.num_children = len(self.data.metadata)
+            self.children.append(node)
+            for key in self.data.read().keys():
+                child = RunNode(self.run, key, '', None, self)
+                self.children.append(child)
+        elif self.data and isinstance(self.data, dict):
```
danielballan, line 88: For `isinstance` checks on core types, prefer `collections.abc.Mapping` and `collections.abc.Iterable` because they will catch duck types as well as the specific types. A concrete example of where we care about this: we want to catch `list` and `tuple`.

bluesky_widgets/_qt/qt_main_window.py
```diff
@@ -53,7 +53,8 @@ def __init__(self, qt_widget, *, show):
         self._qt_center = QWidget(self._qt_window)
 
         self._qt_window.setCentralWidget(self._qt_center)
-        self._qt_window.setWindowTitle(self.qt_widget.model.title)
+        if hasattr(self.qt_widget.model, 'title'):
+            self._qt_window.setWindowTitle(self.qt_widget.model.title)
```
danielballan, line 57: This title injection code was cribbed from napari, and may need a rethink for our use case (ours being a component library vs napari's an extensible application). This is fine for now.

bluesky_widgets/examples/qt_run_tree_view.py
cryos: True, missed this - I will fix it up.

bluesky_widgets/qt/run_tree.py
cryos: Thanks for the explanation, I will fix this up, I was used to using this a lot but I think we were spanning 2 and 3.

bluesky_widgets/qt/run_tree.py
cryos: It would be great to explore, I think you still need the underlying tree model with some knowledge of BlueskyRun, but it would be nice to build more intelligence into how we might update this. Maybe we can improve this in further feature branches, I think I may need a little more help seeing how best to apply what you mention here.

bluesky_widgets/qt/run_tree.py
cryos: I appreciate the nudge, completely agree, and it is much shorter/more readable.

bluesky_widgets/qt/run_tree.py
cryos: Done

cryos:
I think I addressed all review comments, and fixed up flake8 warnings.

PR #24, opened 2020-09-15
Ergonomic entry followup

danielballan:
## User-facing changes

- Initialize the time entry with the 'All' button checked.
- Previously, the radio buttons became un-checked after a refresh. If it button is checked before a refresh it should stay checked.

## Test changes

Added unit tests for absolute and relative time entry, covering:
- Input from the model (e.g. console)
- Input from the datetime pickers
- Input from each of the radio buttons

## Implementation changes
- Connect to `QRadioButton`'s `toggled` signal rather than `clicked`, as recommended by docs. (This makes it easier to test against as well.)
- Use `QObject.blockSignals` to ensure that we never get circular loops where the model updates the view which updates the model which updates the view. Employ a custom context manager to ensure that we never forget to UN-block the signals.
- Refactor to use `datetime.now()` instead of `time.time()` so that we are always handling `datetime` or `timedelta`, not a mix of `datetime` and floating point timestamp.
- Define Ada Lovelace's birthday (which we cheekily use as the lower bound on "All" times) as an absolute `datetime` rather than a `timedelta`.

### DIFF
--- a/bluesky_widgets/_qt/tests/test_qt_search_input.py
+++ b/bluesky_widgets/_qt/tests/test_qt_search_input.py
@@ -77,7 +77,7 @@ def test_time_input_relative_from_model(qtbot, delta, radio_button):
     actual_until = datetime.fromtimestamp(view.until_widget.dateTime().toSecsSinceEpoch())
     assert abs(actual_until - expected_until) < TOLERANCE
 
-    # And a radio button checks.
+    # And a radio button is checked.
     radio_button_widget = getattr(view, radio_button)
     radio_button_widget.isChecked()
 
@@ -107,12 +107,13 @@ def test_time_input_from_radio_buttons(qtbot, delta, radio_button):
     # should update, reevaluating their relative times with respect to the
     # current time.
     time.sleep(1)
-
     model.request_reload()
+
     # Since and until should update with respect to current time.
     new_actual_since = datetime.fromtimestamp(view.since_widget.dateTime().toSecsSinceEpoch())
     new_actual_until = datetime.fromtimestamp(view.until_widget.dateTime().toSecsSinceEpoch())
     assert actual_since != new_actual_since
     assert actual_until != new_actual_until
+
     # And button should still be checked.
     radio_button_widget.isChecked()
--- a/bluesky_widgets/_qt/tests/test_qt_search_input.py
+++ b/bluesky_widgets/_qt/tests/test_qt_search_input.py
@@ -8,7 +8,7 @@ from ...components.search.search_input import SearchInput, LOCAL_TIMEZONE
 
 def as_datetime(qdatetime):
     assert isinstance(qdatetime, QDateTime)
-    return qdatetime.toPyDateTime().replace(tzinfo=LOCAL_TIMEZONE)
+    return QDateTime.toPython(qdatetime).replace(tzinfo=LOCAL_TIMEZONE)
 
 
 def datetime_round_trip(dt):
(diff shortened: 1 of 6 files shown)


bluesky_widgets/_qt/tests/test_qt_search_input.py
```diff
@@ -0,0 +1,139 @@
...
+
+def as_qdatetime(datetime):
+    "Create QDateTime set as specified by datetime."
+    qdatetime = QDateTime()
+    qdatetime.setSecsSinceEpoch(datetime.timestamp())
+    return qdatetime
+
+
+def test_initial_state(qtbot):
+    "Check that 'All' is checked and since/until are set as expected."
```
JulReinhardt, line 17: I'm getting errors when running pytest. 
`recursive dependency involving fixture 'qtbot' detected`
Don't we have to import the qtbot fixture from conftest?

bluesky_widgets/qt/search_input.py
```diff
@@ -212,3 +215,18 @@ def uncheck_radiobuttons(self):
...
+    qobject.blockSignals(True)
+    yield
+    qobject.blockSignals(False)
+
+
+# We need some concrete datetime to show in the "Since" datetime picker when
+# the "All" button is checked. It should be some time old enough that it isn't
+# likely to leave out wanted data. We cheekily choose this birthday as that
+# time.
+ADA_LOVELACE_BIRTHDAY = datetime(1815, 12, 10)
```
JulReinhardt, line 232: I get 1815-12-09-23:52 as all since display. Which I find a bit weird, as it is not n hours different depending on the time zone, but feels a bit random, since it's eg 4:39 pm right now.

JulReinhardt:
Excellent changes Dan. I made some comments about my failing pytest and the weird "All" date display.
I'd suggest to put the refresh button a bit further to the catalog display or rename it to "Refresh catalogs" as I intuitively always expect the datetime to refresh when clicking that. But if I am the only, I am happy leave it.

bluesky_widgets/_qt/tests/test_qt_search_input.py
danielballan: Interesting. What versions of pytest and pytest-qt are you running? It seems to work on Travis-CI and locally for me, so I suspect something is different about our environments.

> Don't we have to import the qtbot fixture from conftest?

That is a very reasonable question, but no, `pytest` is a little _too_ magical at times and one of the things it does is make everything in `conftest.py` automatically available as a test function parameter.

bluesky_widgets/qt/search_input.py
danielballan: Weird. The conversion from `datetime.datetime` to `QDateTime` seems to do something funny:

```py
In [5]: ADA_LOVELACE_BIRTHDAY                                                                                                                                                                 
Out[5]: datetime.datetime(1815, 12, 10, 0, 0)

In [6]: as_qdatetime(ADA_LOVELACE_BIRTHDAY)                                                                                                                                                   
Out[6]: PyQt5.QtCore.QDateTime(1815, 12, 9, 23, 56, 2)
```

Investigating....

bluesky_widgets/_qt/tests/test_qt_search_input.py
JulReinhardt: Whoops...did not have pytest-qt installed. Now it's passing the tests.

bluesky_widgets/qt/search_input.py
danielballan: I finally figured this out and added some additional tests to ensure specifically that we convert from `datetime` to `QDateTime` successfully.

danielballan:
Confusion with `datetime`, `QDateTime`, and time zones stymied me for awhile but I finally got this working as expected. The initial state looks right. Both the model (shown in IPython) and the GUI agree on the date, and there is no unexpected offset in the time fields.

image

Updating the model from the console or the GUI synchronizes as expected.

JulReinhardt (approved):

PR #27, opened 2020-10-01
Do not block on opening catalogs

### LINKED ISSUE #25: Threading issues with mongo server timeout
In the relatively short period of time when the databroker browers was in xicam.gui, I spent some time trying to make it so that searching a catalog for runs didn't block xicam's main thread, and gave some notice to the user in the status bar about what was going on, with a progress bar (what I did needed improvement, for sure.)  Now that it's back to using bluesky-widgets, I'm seeing some issues with the new bluesky-widgets code in xicam.

While waiting for a mongo server timeout to occur, I see two issues:
- [ ] xicam is locked up, which indicates to me that something is blocking the main thread
- [ ] when the timeout does occur and control comes back xicam, the catalog selection dropdown does not re-enable, making it hard to try a different catalog.

danielballan:
Closes #25 

We were blocking the main thread on accessing sub-Catalogs. We already use `create_worker` (a nice wrapper around `QRunnable` that we vendored from napari) to access the items in a Catalog. This PR applies that approach to the initial sub-Catalog access as well, which is when MongoDB client--server connection happens and has the potential to hang for 30 seconds.

I tested this against this configuration file, which points to a nonexistent MongoDB deployment.

```
# ~/.local/share/intake/test_not_connectable.yml 
sources:
  broken_example:
    driver: bluesky-mongo-normalized-catalog
    args:
      metadatastore_db: mongodb://localhost:55555/WRONG_PORT_CANNOT_CONNECT
      asset_registry_db: mongodb://localhost:55555/WRONG_PORT_CANNOT_CONNECT
```

When I select this item in the GUI, the combobox is immediately disabled _but the GUI remains responsive_. After 30 seconds (pymongo's default [server selection timeout](https://api.mongodb.com/python/current/api/pymongo/mongo_client.html#pymongo.mongo_client.MongoClient)) the traceback is logged, the combox is re-enabled and reset to a blank value.

It would be nice if we could back out before waiting for the timeout---that is, to give the option to cancel the operation before it succeeds for times out. Cancellation is generally difficult to get right, so I would like to punt that to a future PR.

As is, this at least _does not_ lock up the application and _does_ leave it in a usable state after the failure.

Needs a unit test

### DIFF
--- a/bluesky_widgets/components/search/searches.py
+++ b/bluesky_widgets/components/search/searches.py
@@ -127,9 +127,9 @@ class Search:
             old = self._root_catalog
         new = old[name]
 
-        # Touch an attribute that will trigger a connection attempt. (It's here
+        # Access an entry to trigger a connection attempt. (It's here
         # that an error would be raised if, say, a database is unreachable.)
-        new.metadata
+        next(new.items())
 
         # If we get this far, it worked.
         self._subcatalogs.append(new)
--- a/bluesky_widgets/qt/searches.py
+++ b/bluesky_widgets/qt/searches.py
@@ -84,15 +84,16 @@ class QtSearch(QWidget):
             name = names[index]
             selector.setEnabled(False)
 
-            def on_failure():
-                logger.exception("Failed to select %r", name)
+            def on_errored(err):
+                logger.exception("Failed to select %r", name, exc_info=err)
+                # Reset the combobox selection to an empty value and enable it.
                 selector.setCurrentIndex(-1)
                 selector.setEnabled(True)
 
             worker = create_worker(
                 self.model.enter,
                 name,
-                _connect={"errored": on_failure}
+                _connect={"errored": on_errored}
             )
 
         selector.activated.connect(on_selection)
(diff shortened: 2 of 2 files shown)


danielballan:
P.S. This work reinforced my view that the `Search` abstraction, which currently includes sub-Catalog combo boxes and forms/results for searching a Catalog of Runs, needs to be refactored. Not a huge priority, but should make this easier to look at.

dylanmcreynolds:
I pull this down, and it seems to have to have fixed the locking problem. Thanks!

danielballan:
I have let this languish because I have reconsidered this statement above:

> It would be nice if we could back out before waiting for the timeout---that is, to give the option to cancel the operation before it succeeds for times out. Cancellation is generally difficult to get right, so I would like to punt that to a future PR.

If we refactor the model to accept _absolute_ sub-catalog paths, rather than "Go down one level" or "Go up one level," it will be perfectly possible to do cancellation. I think it's important to not make the user wait for the timeout.

danielballan:
This is waiting on #62, in which I can easily implement the proposal in my comment above, and then this will work nicely.

PR #34, opened 2020-11-10
Remove some nesting

gwbischof:


danielballan:
I think `views` adds verbosity without much value. I would rather leave the views (`qt`, `jupyter`) as top-level subpackages under `bluesky_widgets` just for the sake of brevity.

bluesky_widgets/qt/search.py
```diff
@@ -0,0 +1,5 @@
+from ._run_tree import *
```
danielballan, line 1: The `run_tree` is its own thing, not related to search. Should remain a public module under `qt`.

danielballan (commented):
This certainly makes the imports simpler and easier to remember.

danielballan:
All builds passed except the `flake8` one. Ran `black` and added some `flake8` `noqa: ...` comments. No changed to the ast. Merging.

PR #35, opened 2020-11-10
QtSearchList

gwbischof:


gwbischof:
This is showing so many changes because I made this PR on top of the other PR.

gwbischof:
Can we check the type of the model that gets passed to the View and raise if it is the wrong type?

gwbischof:
Also can we always name the Views that match a model following some standard?
For example if the model is: SearchList
Then the Qt view would be: QtSearchList
And the Juptyer view would be: JupyterSearchList
And the Xicam view would be: XicamSearchList?

danielballan:
> Can we check the type of the model that gets passed to the View and raise if it is the wrong type?

Yes and no. Yes, in the medium-term we should explore [static type checking](http://mypy-lang.org/), which is being used increasingly in Python. (This is another place where we can learn by studying napari, among other things.) No, a literal `isinstance(view, ExpectedModelClass)` is too limiting. Anything that duck-types as a model (provides the right methods) should be accepted so that the code can be reused in ways not foreseen by us. 

> Also can we always name the Views that match a model following some standard?

_All_ of the thumbs up for this.

Regarding the relevant changes in the PR (so, just 244056a) I see something I like and something I would change:

* I like moving the "processing" code to a model and calling `self.model.handle_click()` in the view. As I wrote it, the example was cheating by doing "processing" in the view, and that makes it is poor example.
* I would prefer to move this into a new, separate model and view, defined in `qt_serach.py`, rather embedding the assumption that a search list will always have _one button_ in `SearchList` and `QtSearchList`. By defining these things in `qt_search.py`, we will show clearly where to put the code that _does_ something ("Just edit the body of `handle_click()`!") and we will lay track for having more than one button, or other interactions. For example:
  * A pane that shows a preview when a result in the table of results is first clicked (no need to click a button as well)
  * A button for "Process data" and a button for "View data" and button for "Export data" which might open a dialog box, etc.

  All of which to say: the `SearchList` as it is represents an extensible component, and things like `handle_click()` belong on models that extend it in a particular way. Some of these might work their way into "core" (`SearchListWithOneButton`? Who knows!) but until we have clear feedback on what the commonly-useful bits are, I'd like to experiment in the examples with objects at this level.

gwbischof:
I'll look into static type checking for a future PR. I like that idea! An Error at import time would be nice.

Makes sense to subclass SearchList to add a button. But we will have to make a subclass each time we use it because each application with want to do something different with handleClick. What if we make this an Abstract Base Class in core, then the example application sub-classes and overrides? Also we pass a function as an argument to set handleClick.

Maybe we use this pattern? https://github.com/bluesky/bluesky-kafka/blob/master/bluesky_kafka/__init__.py#L405

gwbischof:
For this PR I'll start with making a subclass of SearchList for the version with a button.

danielballan:
> What if we make this an Abstract Base Class in core, then the example application sub-classes and overrides? Also we pass a function as an argument to set `handleClick`.

If we find that it's a common use case to want to add one button to `SearchList`, then adding something like makes sense. What I was trying to get at above is that it's too early to know if that's actually a common use case, or whether every application will add very different button sets.

gwbischof:
Maybe I'll add it anyway so that we have an example of doing this?

danielballan:
:+1: for adding it in `qt_search.py` so we have an easy-to-find example of how to _properly_ extended `SearchList` with `handle_click` rather than my current anti-exmaple.

gwbischof:
Rebased

danielballan (commented):
It's premature to generalize `SearchListWithButton` as a reusable component. Please restrict the changes here to `qt_search.py`, merging `SearchListWithButton` and `SearhListWithButtonExample` into one class defined there, without the injection-via-`__init__`-parameter. I appreciate the potential future utility of generalizing like this, but it's too early to know what the right extension points are.

gwbischof:
But I wanted to show an example that included how to write the library code, and how to extend it in the application code. Is there a way that I can still do this?

danielballan:
I'm feeling stubborn on this point because I really think this is a case where a little duplication is better than an additional abstraction. ("Premature abstraction is the root of all evil," applies.) It's OK to have the same code repeated in a couple examples, especially at this stage.

bluesky_widgets/examples/qt_search.py
```diff
@@ -9,13 +9,27 @@
...
+            print(
+                f"Processing Run {uid[:8]} (scan_id={run.metadata['start']['scan_id']})"
+            )
 
 
-class SearchesWidget(QWidget):
+class QtSearchListWithButton(QWidget):
     """
-    Combine the QtSearches widget with a button that processes selected Runs.
+    Combine the QtSearcheslist widget with a button that processes selected Runs.
```
danielballan, line 31: Add a blank line so this renders nicely in Sphinx.

bluesky_widgets/examples/qt_search.py
```diff
@@ -9,13 +9,27 @@
 from bluesky_widgets.qt.search import QtSearches
 from bluesky_widgets.examples.utils.generate_msgpack_data import get_catalog
 from bluesky_widgets.examples.utils.add_search_mixin import columns
+from qtpy.QtWidgets import QWidget, QPushButton, QVBoxLayout
 
-from qtpy.QtWidgets import QPushButton, QVBoxLayout, QWidget
+
+class SearchListWithButton(SearchList):
+    """
+    Specialize the bluesky-widget model for the application,
```
danielballan, line 17: Is it possible to state this as

```
<One-line summary>

<Further explanation>
```

so it will render nicely in sphinx?

gwbischof: Updated

danielballan (approved):

danielballan:
Tested interactively locally and ran `pytest` locally. Merging without CI because Travis is backed up.

PR #38, opened 2020-11-13
Napari databroker

gwbischof:
An app that combines has the napari viewer and databroker search.


danielballan:
Pushed commit fixing style. Tested locally because Travis is out of credits.

PR #39, opened 2020-11-16
Interactive tree

cryos:


cryos:
I think your changes look good, please merge if you think the rest looks good. This is a huge improvement over my initial version, and I think the example model is great as an extremely simple one for others to see the concepts being used. The API changes are great setting the run as a simple property.

PR #40, opened 2020-11-17
Make Search tabs not closable.

danielballan:
Search tabs can be added or removed from the model (e.g. console).
Currently, they can be removed (closed) from the UI but not added. This
seems bad, as you could accidentally close a tab and have no way to get
it back. For now, I think it's better to make the tabs not closable from
the UI.

### DIFF
--- a/bluesky_widgets/qt/_searches.py
+++ b/bluesky_widgets/qt/_searches.py
@@ -155,15 +155,12 @@ class QtSearches(QTabWidget):
         self.insertTab(event.index, tab, f"Search {event.item.name}")
 
     def on_removed(self, event):
-        if event.item in self._tabs:
-            # A Search has been removed from the SearchList model.
-            # Close the associated tab and clean up the associated state.
-            widget = self._tabs[event.item]
-            index = self.indexOf(widget)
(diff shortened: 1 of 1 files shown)


danielballan:
I pushed a second commit that simplifies the tab-closing logic (similar to how I just implemented it in the Viewer tabs) which we be useful if/when we add UI for adding search tabs someday and reinstate the ability to close them from UI. Confirmed that closing tabs from the model side works as before, but now implemented more cleanly.

```py
In [2]: from bluesky_widgets.examples.qt_search import ExampleApp

In [3]: app = ExampleApp()

In [4]: del app.searches[0]  # tab closes
```

cryos (approved):
The position is perhaps a little sad, but it looks like the most reasonable thing to do in the short term. It may want tweaking in the future to enable the caller to offer closable tabs - the larger application the widget is embedded within may offer a way of adding tabs even though the individual widget does not.

danielballan:
Agreed. I don't have any clear User Stores in my head yet related to why you would want to be adding or removing tabs, but it seems potentially useful. Worth revisiting once we have some Stories to target.

PR #46, opened 2020-11-24
Plots

danielballan:
# Goals

This is intended to be successor to both:

* bluesky's plotting callbacks (e.g. `bluesky.callbacks.mpl_plotting.LivePlot`) which were implemented in 2015 in a rush and need to be revisited with more thought to design
* bluesky's "Best-Effort Callback" (BEC), a system for automatically inferring what to plot based on "hints" in the documents, and then generating those plots

Specifically, we want to:

1. **Support a range of visualization frameworks.** Support Jupyter and Qt via matplotlib, just as the old LivePlot and BEC do. Grow to support other visualization frameworks beyond matplotlib, especially frameworks that render in the browser on the client side (e.g. Plotly, Bokeh, bqplot, vega / d3, [various react chart libraries](https://blog.bitsrc.io/top-5-react-chart-libraries-for-2020-194acc5a15cf)).
2. **Support "headlines" rendering.** If we build tools that produce useful figures during an experiment, we should be able to reuse them to generate thumbnail images.
3. **Do not restrict visualization to a "Figure".** Integrate more deeply with GUI frameworks. Incorporate other widgets, starting with a simple "tabbed container" of figures and extending to other widgets such as input fields and sliders that enable the richer user interaction necessary for things like an "image stack viewer".
4. **Support "streaming" and "prompt" use cases.** "Prompt" means generating a visualization as soon as a Run completes, but not during a Run. It is easier to implement than full streaming, and it should be available for users who do not need streaming. (LivePlot and BEC gave no such option; you could not opt out of the complexity of streaming.)
5. **Support plotting multiple Runs, interleaved.** This is a narrow but important technical limitation of LivePlot: you can't ask it to plot Run X if it's still in the middle of consuming data from Run Y. This means you cannot add a saved scan while a live scan is streaming in.
6. **Retain a connection between a Run and the plot elements that relate to it.** This enables features like changing the style of a line based on its Scan ID, or changing the color of a live-updating line when the associated scan completes, or plotting only the N most recent runs in a streaming fashion. Again, this is a limitation of LivePlot: it does not retain knowledge of a mapping between Runs and lines.
7. **Support user-specified nontrivial transformations between the data as it was received (from RunEngine or databroker) and the plot.** LivePlot only supports directly plotting columns of the data, given field names x and y. Users should be allowed to inject arbitrary computation here. Currently, this is possible but not straightforward. It requires the author to subclass `LivePlot` and understand much of its internal workings. It also requires the author to work with documents, not xarrays/dataframes/ndarrays. See for example BMM's [DerivedPlot](https://github.com/NSLS-II-BMM/profile_collection/blob/0c35e1c8d6428f25cc4ca0dd875bd010592e2713/startup/save/65-derivedplot.save#L68), created with some guidance from us.
8. **Support user-specified rules for deciding what plots to make.** BEC hard-codes logic for parsing the "hints" in the Run Start document to decide what plot types would be useful. It is one giant monolith. There should be a reusable mid-level abstraction that enables the user to configure how to make these decisions based on custom metadata, "projections", "hints", or whatever they like.

# Prior and related work

This incorporates lessons learned from several abandoned attempts to get this right:

* https://github.com/NSLS-II/album (archived 2015)
* https://github.com/NSLS-II/databroker-browser (archived 2018)
* https://github.com/bluesky/bluesky-browser (archived 2019)
* https://github.com/bluesky/bluesky-mpl/ (matplotlib-related remnants of bluesky-browser, archived today)

as well as from discussion in the Pilot calls and weekly ongoing conversations with the Xi-CAM team doing related work around "projections" and their concept of "intents".

# Initial use cases

The following use cases were targeted in this PR, to validate the design.

* Plot a given y vs x for the N most recent scans streamed from the RunEngine together on one set of axes, along with some saved scans loaded from databroker. Related issues:
    * https://github.com/bluesky/bluesky-mpl/issues/4
    * https://github.com/bluesky/bluesky/issues/1270
* When multiple figures are created, placed them in a tabbed contained.
    * https://github.com/bluesky/bluesky/issues/1269
* View images.

# Screenshots

image
image

image

# How this PR addresses the goals

1. **Support a range of frameworks.** This relies on custom, framework-agnostic models of a figure and its contents, which is then fed to a given view to render it.

    ```py
    # Define a framework-agnostic model.
    from bluesky_widgets.models.plot_specs import FigureSpec, AxesSpec, LineSpec  # to name just a few...
    line = LineSpec(...)
    axes = AxesSpec(lines=[lines], ...)
    figure = FigureSpec(axes=(axes,), ...)

    # Feed it to a framework-coupled view.
    from bluesky_widgets.qt.figures import QtFigure
    view = QtFigure(figure)

    from bluesky_widgets.jupyter.figures import JupyterFigure
    view = JupyterFigure(figure)

    # Modifying the model instantly updates the view.
    line.style.update(color="red")  # Line color updates in view!
    another_line = LineSpec(...)
    axes.lines.append(another_line)  # New line is added to view!
    figure.title = "New title"  # Figure title updates in view!
    ```
    These are inherently limiting: you can only specify the things that we let you specify in that "framework-agnostic" model, which is of course a framework of its own! With great generality comes great constraint. But, there is an escape hatch. You can always access the full power of a given framework by accessing the view. For example, I could draw a [quiver plot](https://matplotlib.org/api/_as_gen/matplotlib.axes.Axes.quiver.html) even though the model has no such concept.
    ```py
    # Escape to matplotlib!
    fig = view.figure  # the actual matplotlib.figure.Figure instance
    fig.axes[0].quiver.(...)
    ```
   Thus, you can enjoy our framework's benefits without being totally confined by its limitations.

   A _list_ of figures is represented by a the view as a tabbed container.

    ```py
    model = FigureSpecList([figure1, figure2])
    view = JupyterFigures(model)  # ipywidgets.widgets.Tab
    view = QtFigures(model)   # qtpy.QtWidgets.QTabWidget
    ```

    The models are implemented using the same pattern and signaling system we use in bluesky-live and elsewhere in bluesky-widgets. They make it possible for users to work from the graphical user interface or from the console from a mixture of both.
2. **Support "headlines" rendering.** This falls out naturally from the design in (1). Imagine a new view type `HeadlessFigures` that has no graphical UI but implements a method `view.save_all_figures_to_jpeg("path/to/directory")`.
3. **Do not restrict visualization to a "Figure".** This PR includes the model `FigureSpecList`, which `QtFigures` and `JupyterFigures` represent as a tabbed contained of `QtFigure` and `JupyterFigure` widgets respectively. In the future, we can add framework-agnostic models for slider bars, input boxes, or other elements, and each view can implement these within its respective framework.
4. **Support "streaming" and "prompt" use cases.** Writing custom _streaming_ visualization logic requires handling various types of events and contingencies. Writing custom _prompt_ visualization logic allows the user to make more assumptions; they know they have access to all the data and they don't need to handle updates. This PR includes a prototype of both streaming and prompt plot types.
5. **Support plotting multiple Runs with interleaved data.** LivePlot operated on a flat stream of documents, which made it hard to support interleaved Runs. Now, `event_model.RunRouter` and `bluesky_live.BlueskyRun` work together to handle this so that the plotting code can just operate on `BlueskyRun`s. The simplest illustration is:

    ```py
    from bluesky_widgets.utils.streaming import stream_documents_into_runs

    runs = []
    RE.subscribe(stream_documents_into_runs(runs.append))
    # As documents stream in, BlueskyRuns will get appended to runs.
    ```

6. **Retain a connection between a Run and the plot elements that relate to it.** A given `FigureSpec` or `FigureSpecList[FigureSpec]` is managed by some higher-level model ("plot builder") which also keeps a `BlueskyRun` or `RunList[BlueskyRun]` and keeps track of the correspondence between figures (or axes or lines....) and Runs. An example of a plot-builder model is `RecentLines`.

    ```py
    model = RecentLines(3, "motor", "det")
    model.figure  # FigureSpec
    model.runs   # currently viewed runs
    ```
    Deleting an item from `model.runs` removes the corresponding line for `model.figure`. As above, the figure can be viewed like

    ```py
    view = QtFigure(model.figure)
    ```

    and the view will update to reflect changes to the model.

7. **Support user-specified nontrivial transformations between the data as it was received (from RunEngine or databroker) and the plot.** There is more thought needed on how best to do this, but `RecentLines` and `Image` plot-builder enable it like so:

    ```py
    class RecentLines:                                                              
          """                                                                         
          Plot y vs x for the last N runs.                                            
                                                                                  
          Parameters                                                                  
          ----------                                                                  
          max_runs : int                                                              
              Number of lines to show at once                                         
          x : string                                                                  
              Field name                                                              
          y : string                                                                  
              Field name                                                              
          stream_name : string, optional                                              
              Stream where fields x and y are found. Default is "primary".            
          func : callable, optional                                                   
              Expected signature::                                                    
                                                                                  
                  func(run: BlueskyRun, stream_name: str, x: str, y: str) -> x: Array, y: Array
                                                                                  
              Default::                                                               
                                                                                  
                  def func(run, stream_name, x, y):                                   
                      ds = run[stream_name].to_dask()                                 
                      return ds[x], ds[y]                                             
                                                                                  
          axes : AxesSpec, optional                                                   
              If None, an axes and figure are created with default labels and titles  
              derived from the ``x`` and ``y`` parameters.  
        ....
        """
    ...

    class Image:                                                                    
         """                          
         Plot an image from a Run.

         By default, higher-dimensional data is handled by repeatedly
         averaging over the leading dimension until there are only two dimensions.      
                                                                                  
          Parameters                                                                  
          ----------                                                  
          field : string                                                              
              Field name ("data key") for this image                                  
          stream_name : string, optional                                              
              Stream where fields x and y are found. Default is "primary".            
          func : callable, optional                                                   
              Expected signature::                                                    
                                                                                  
                  func(run: BlueskyRun, stream_name: str, x: str, y: str) -> x: Array, y: Array
                                                                                  
              Default::                                                               
                                                                                  
                  def func(run, field):                                               
                      ds = run[stream_name].to_dask()                                 
                      data = ds[field].data                                           
                      # Reduce the data until it is 2D by repeatedly averaging over   
                      # the leading dimension until there only two dimensions.                   
                      while data.ndim > 2:                                            
                          data = data.mean(0)                                          
                      return data              
          axes : AxesSpec, optional                                                   
              If None, an axes and figure are created with default labels and titles.
          ...
        """
        ...
    ```

8. **Support user-specified rules for deciding what plots to make.** Some ideas on this have been discussed, but this problem is not yet solved and is left out of scope for this PR.

This PR does _not_ yet achieve feature parity with Best-Effort Callback. Progress toward that is tracked via a GitHub Issues label: https://github.com/bluesky/bluesky-widgets/issues?q=is%3Aissue+is%3Aopen+label%3ABEC-feature-parity

## Overview of code changes

* `bluesky_widgets.models.plot_specs` implements the "framework-agnostic" models of a figure and its contents.
* `bluesky_widgets.models.plot_builders` implements models that join `FigureSpec`(s) to `BlueskyRun`(s) and encode rules for drawing and updating figures from runs and tracking the connection between plot objects and runs.
* The views are implemented in `bluesky_widgets.qt.figures` and `bluesky_widgets.jupyter.figures`. Because these both happen to be matplotlib-based they can share a lot of code, which is in `bluesky_widgets._matplotlib_axes`. Any future matplotlib-based views (like a matplotlib `HeadlessFigures`) can reuse this as well. Other views (like `PlotlyFigures`) will not.
* The support code in `bluesky.utils.event` and `bluesky.utils.list` has been copied into the library `bluesky-live`. We now import it from there instead of maintaining a separate copy here.
* Various examples for Python, IPython, and Jupyter have been added to `bluesky_widgets.examples`.
* To illustrate streaming from a separate process, `bluesky_widgets.examples.utils.stream_data` was added. This is largely copy pasted from `bluesky-browser` which had something similar. Streaming from 0MQ into a Jupyter kernel or a Qt application requires special handling of concurrency tailored to each case. These are implemented in `bluesky.qt.zmq_dispatcher` and `bluesky.jupyter.zmq_dispatcher`. They are sketchy (specifics are noted in code comments) but I would like to punt refinements to a future PR. They are used in `bluesky.examples.advanced.qt_viewer_with_search` and `bluesky/examples/Streaming.ipynb`. (Yes, we should support Kafka too. The nice thing of 0MQ for examples is that it can be run with user-level permissions.)
* Some modules were rearranged. I consolidated `_jupyter` and `jupyter` into just `jupyter`, and the same for `_qt` and `qt`. I deleted the (empty) `xi_cam` subpackage entirely because we have [demonstrated](https://github.com/Xi-CAM/Xi-cam/pull/14) that we can integrate with Xi-CAM without doing any special wrapping on this side.

## Documentation and test coverage

There are thorough docstrings, with examples, and rendered API Reference documentation. More examples should added after the design anneals a bit more.

There are a dozen or so unit tests covering the new functionality, enough to and validate that the design is possible to write good unit tests for. More are needed but should probably not be written until the design anneals a bit more.


bsobhani:
It is neat how you are able to get these into Jupyter. I haven't used Jupyter much, but I would imagine that normally you would look at the different aspects of the data by repeatedly changing/running a cell. Widgets may help make the user experience more interactive.

PR #53, opened 2020-11-30
Test short title syncing

AbbyGi:


bluesky_widgets/qt/_tests/test_figures.py
```diff
@@ -56,6 +56,25 @@ def test_figure_title_syncing(qtbot):
     assert view.figure._suptitle.get_text() == expected
 
 
+def test_short_title_syncing(qtbot):
+    model = func(run)
+    figures = FigureSpecList([model])
+    view = QtFigures(figures)
+    actual_title = view.figures[next(iter(view.figures))].figure._suptitle.get_text()
```
danielballan, line 63: The `next(iter(...))` trick is useful for collections that are not indexable, but in this case I think we just have a `list`, so the simpler `view.figures[0]` will work.

bluesky_widgets/qt/_tests/test_figures.py
```diff
@@ -56,6 +56,25 @@ def test_figure_title_syncing(qtbot):
...
+    assert view.tabText(0) == actual_title
+    expected_short_title = "new short title"
+    model.short_title = expected_short_title
+    assert model.short_title == expected_short_title
+    actual_title = view.figures[next(iter(view.figures))].figure._suptitle.get_text()
```
danielballan, line 68: Ditto here

danielballan (commented):
Looks good. Could you also add a second, separate test function exercising the case where the `model` has a non-null `short_title` when it is first passed to `QtFigures`?

bluesky_widgets/qt/_tests/test_figures.py
danielballan: As you reminded me offline, this is a `DictView`, not a `list`. The dict is keyed on `FigureSpec.uuid`, so the way to access an item here is `view.figures[model.uuid]`.

danielballan (approved):

PR #57, opened 2020-12-01
Refactor pinned_runs concept.

danielballan:
Instead of keeping two separate lists of runs, `pinned_runs` and `runs`,
keep just `runs` and track "pinned" as a set of runs (their `uid`s to be
precise) in that list.

This makes the implementation simpler and avoids the confusing question
whether a run could be in both lists at once. As before, a pinned run
can be added like:

```py
model.add_run(run, pinned=True)
```

and, also as before, a pinned run can be removed the same as an
un-pinned one, by calling

```py
model.discard_run(run)  # doesn't matter if it's pinned or unpinned
```

or removing it from the list by position:

```py
del model.runs[3]  # doesn't matter if it's pinned or unpinned
```

You can effectively "un-pin" a run by removing and then re-adding it.

```py
model.add_run(run, pinned=True)
model.discard_run(run)
model.add_run(run)  # re-add, not pinned
```

(I don't think un-pinning deserves a special method, especially because
it's not clear yet that it will be a common thing to do.)

The set of currently pinned runs (their uids) can be viewed as:

```py
model.pinned  # returns a frozenset of run uids
```

### DIFF
--- a/bluesky_widgets/examples/AutoRecentLines.ipynb
+++ b/bluesky_widgets/examples/AutoRecentLines.ipynb
@@ -103,7 +103,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "model.pinned_runs.append(scans[-1])"
+    "model.add_run(scans[-1], pinned=True)"
    ]
   },
   {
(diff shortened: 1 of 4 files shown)


tacaswell (approved):

jklynch (changes requested):
I would be so happy to see a test or two for the pinned runs. Maybe I just couldn't find them?

jklynch (approved):

PR #66, opened 2020-12-04
Support Python expressions in plot builders

danielballan:
Two major changes here:

* `RecentLines` expects a _list_ of y variables to plot on the same axes, instead of just one y variable.
* Both `RecentLines` and `Image` accept **any valid Python expression**. The expression is evaluated in a namespace that contains the `DataArrays` in the primary stream (e.g. `"motor"`), the `Datasets` of the streams themselves (e.g. `"primary"`, `"baseline"`), the run (`"run"`), and any user-provided tokens. It also includes all numpy objects like `log` and `pi`. The namespace can be extended by the user as illustrated below to inject custom functions or values. In this way, simple data transformations are easy --- `"log(I0)" --- and complex operations like interpolating between streams are possible. Because the input is text it is easy to image providing these expressions via GUI input. It's also interesting to consider "saving favorite plots" by writing these strings into a state store.

This enables usages like...

Plot a mathematical transformation of the columns using any object in numpy.

```py
    >>> model = RecentLines(3, "abs(motor)", ["-log(det)"])
    >>> model = RecentLines(3, "abs(motor)", ["pi * det"])
    >>> model = RecentLines(3, "abs(motor)", ["sqrt(det)"])
```

Plot multiple lines.

```py
    >>> model = RecentLines(3, "motor", ["log(I0/It)", "log(I0)", "log(It)"])
```

Plot every tenth point.

```py
    >>> model = RecentLines(3, "motor", ["intesnity[::10]"])
```

Access data outside the "primary" stream, such as a stream name "baseline".

```py
    >>> model = RecentLines(3, "motor", ["intensity/baseline['intensity'][0]"])
```

As shown, objects from numpy can be used in expressions. You may define
additional words, such as `sg` for a Savitzky-Golay smoothing filter,
by passing it a dict mapping the new word to the new object.

```py
    >>> import scipy.signal
    >>> namespace = {"sg": scipy.signal.savgol_filter}
    >>> model = RecentLines(3, "motor", ["sg(intensity, 5, 2)"],
    ...                     namespace=namespace)
```

Custom, user-defined objects may be added in the same way.

```py
    >>> def f(data):
    ...     ...
    ...
    >>> namespace = {"my_custom_function": f}
    >>> model = RecentLines(3, "motor", ["my_custom_function(intensity)"],
    ...                     namespace=namespace)
```

The same works on Images.

```py
    >>> model = Images("my_filter(ccd)", namespace={"my_filter", f})
```


danielballan:
Nearly-useful example with real (public, CC0-licensed) data from BMM:

```py
In [1]: from bluesky_widgets.models.plot_builders import RecentLines

In [2]: from bluesky_widgets.qt.figures import QtFigure

In [3]: import databroker

In [4]: run = databroker.catalog["BMM"][-1]

In [5]: model = RecentLines(3, "dcm_energy", ["log(I0/It)", "log(I0)", "log(It)"])

In [6]: view = QtFigure(model.figure)

In [7]: view.show()
```
image

danielballan:
TO DO:

* [ ] Support pytest-like magical argument inspection, such that `RecentLines(3, lambda It, I0: It/I0)` is allowed.
* [ ] When adding field names to the eval namespace, only add fields from the streams in `needs_streams` and treat it as a precedence order to resolve what to do when a field appears in multiple streams.
* [ ] Take control of the `AxesSpec` title but do not touch with `FigureSpec`. Giving `axes` to `RecentLines` is giving it control of title / labels, but it should not mess with the figure that those `axes` are embedded in.

danielballan:
I intend to merge this (or ask someone to...) by the end of today to keep things rolling. Nothing is in stone.

AbbyGi (approved):

cryos (approved):
Looks good to me

Issue #52, opened 2020-11-30
In Jupyter, make figures created during plan execution display immediately.


danielballan:
In Jupyter, any figure widgets created during plan execution will _display_ but will not receive data until the cell is done executing. Thereafter, any _changes_ to the figure (e.g. additional plans executed that add lines to a plot) will live-update during cell execution as desired.

It seems we need a way to tell ipywidgets to service the new figure widget.

danielballan:
We have a workaround in #54, but it hammers `figure.canvas.draw()` more than it should. I am leaving this open to track any progress on a better solution.

danielballan:
In #68 we stop hammering `draw()`, and the reliability seems good. However, we have to constrain ourselves to avoid any methods that require a draw, like `fig.tight_layout()` or `axes.legend(loc="best")`. We can relax that constraint by routing the signals through traitlets. This might be straightforward but it would take some time to test it thoroughly, so I'll leave that for a separate PR.

PR #68, opened 2020-12-07
More mpl jupyter fixes

danielballan:
This was the result of hours of interactive testing with @tacaswell. The three example notebooks test:

* Streaming from RE (in process) to `JupyterFigure`
* Streaming from RE (in process) to `JupyterFigures`
* Stream via 0MQ (with RE in a separate process) to `JupyterFigures`

and all three work reliably. This does _not_ address potential issues with dropped PNG diffs (i.e. the lack of keyframes in ipympl) but it does address the major rendering issues in current `master` whereby `draw()` was sometimes being called from more than one thread.


danielballan:
We can do better still (see #52) but this is a significant improvement, so I'm going to merge it.

PR #71, opened 2020-12-07
Kafka figures

gwbischof:
Make an example that subscribes to a kafka topic and generates figures.


gwbischof:
I'm not sure about the right way to hook up the RunRouter yet. I need to investigate.

danielballan:
All the RunRouter details should be handled by `stream_documents_to_run`. Something like this should do:

```py
from bluesky_widgets.models.utils import run_is_live_and_not_completed
...

def export_thumbnails_when_complete(run):
    "Give a BlueskyRun, export thumbnail(s) to a directory when it completes."
    model =...
    model.add_run(run)
    view = HeadlessFigures(model)
    directory = "the uid or something"
    # If the Run is already done by the time we got it, export now.
    # Otherwise, schedule it to export whenever it finishes.
    if run_is_live_and_not_completed(run):
        run.events.completed.connect(view.export_all(directory))
    else:
        view.export_all(directory)

# stream_documents_into_runs organizes the flat stream of (name, doc) pairs
# from the dispatcher into BlueskyRun objects. When the dispatcher emits a 'start'
# document, a BlueskyRun will be created and immediately passed to
# export_thumbnails_when_complete.
# Future documents belonging to that run will be added to that BlueskyRun's internal DocumentCache.
# Signals on the run like `run.events.new_data` and `run.events.completed` can be
# used to detect updates. At any point, the (partial) data can be accessed
# as documents or as xarrays.
dispathcher.subscribe(stream_documents_into_runs(export_thumbnails_when_complete))
```

bluesky_widgets/headless/figures.py
```diff
@@ -62,6 +62,12 @@ def _on_figure_removed(self, event):
         figure.close_figure()
         del self._figures[figure_spec.uuid]
 
+    def close_figures(self):
+        for figure in self._figures.values():
+            _close_figure(figure)
```
danielballan, line 67: The keys in this dict are `HeadlessFigure` instances, not `matplotlib.figure.Figure` instances, so this should be `figure.close_figure()`.

bluesky_widgets/examples/kafka_figures.py
```diff
@@ -0,0 +1,68 @@
...
+from bluesky_widgets.models.utils import run_is_live_and_not_completed
+
+
+def export_thumbnails_when_complete(run):
+    "Given a BlueskyRun, export thumbnail(s) to a directory when it completes."
+    model = AutoRecentLines(3)
+    model.add_run(run)
+    view = HeadlessFigures(model.figures)
+
+    home = str(Path.home())
```
danielballan, line 29: For examples it's politer to write into `/tmp` so that the stuff you generate doesn't pollute the caller's home directory and so that it doesn't risk clobbering any of their existing files. (Even if the risk of an actual name collision is negligible in a given case, when software does this I wonder what _other_ files in my `$HOME` they might decide to write to, and I worry....)

In this example we make a temporary directory with an random unique name under `/tmp` and then print the files' location.

https://github.com/bluesky/bluesky-widgets/blob/7dbe644d5cc2c2fdb7831650fe76d6025aadcbe6/bluesky_widgets/examples/headless_figures.py#L39-L41

Adding a `print` inside `export` might be useful here too, because they would give feedback on when it has run. (Yesterday we had to watch `ls` to see that.)

bluesky_widgets/examples/kafka_figures.py
```diff
@@ -0,0 +1,68 @@
...
+    base_directory = os.path.join(home, "bluesky_widgets_test")
+    if not os.path.exists(base_directory):
+        os.makedirs(base_directory)
+    uid = run.metadata['start']['uid']
+    directory = os.path.join(base_directory, uid)
+    if not os.path.exists(directory):
+        os.makedirs(directory)
```
danielballan, line 36: `os.makedirs(directory, exist_ok=True)` will let you avoid the `if ...`. Also, You can make nested directories this way ("make dirS") so no need to make the `base_directory` separately.

bluesky_widgets/examples/kafka_figures.py
```diff
@@ -0,0 +1,68 @@
...
+    # If the Run is already done by the time we got it, export now.
+    # Otherwise, schedule it to export whenever it finishes.
+    def export(event):
+        view.export_all(directory)
+        view.close()
+
+    if run_is_live_and_not_completed(run):
+        run.events.completed.connect(export)
+    else:
+        view.export_all(directory)
```
danielballan, line 47: Might as well reuse `export` here too.

danielballan (commented):
Seems close to ready to merge as a good first example. I left a couple comments.

gwbischof:
Updated all of the suggestions, no more warnings!

danielballan:
Added some commits to clean up after the renaming of `bluesky_widgets.examples.utils.stream_data` and other minor stuff. Merging.

PR #73, opened 2020-12-08
Add RasteredImage (née Grid) functionality

AbbyGi:


danielballan:
These parameters in `__init__` are just matplotlib parameters:

```
    clim : tuple, optional
       The color limits

    cmap : str or colormap, optional
       The color map to use

    extent : scalars (left, right, bottom, top), optional
       Passed through to :meth:`matplotlib.axes.Axes.imshow`

    aspect : str or float, optional
       Passed through to :meth:`matplotlib.axes.Axes.imshow`
```

which we can capture in `__init__` and just pass through via `style` as in:

```py
image = ImageSpec(func, self.run, label=self.field, style=...)
```

I'm not sure it's quite right lump these parameters into "style" --- they are more fundamental that something like line color --- but it seems a reasonable and easy first pass.

These two are used by `LiveGrid` to flip the image vertically or horizontally by flipping the axes limits.

```
    x_positive: string, optional
        Defines the positive direction of the x axis, takes the values 'right'
        (default) or 'left'.

    y_positive: string, optional
        Defines the positive direction of the y axis, takes the values 'up'
        (default) or 'down'.
```

`AxesSpec` will have grow new properties and events for `x_limits` and `y_limits`, and then this code

```
        # make sure the 'positive direction' of the axes matches what
        # is defined in axes_positive
        xmin, xmax = self.ax.get_xlim()
        if ((xmin > xmax and self.x_positive == 'right') or
                (xmax > xmin and self.x_positive == 'left')):
            self.ax.set_xlim(xmax, xmin)
        elif ((xmax >= xmin and self.x_positive == 'right') or
                (xmin >= xmax and self.x_positive == 'left')):
            self.ax.set_xlim(xmin, xmax)
        else:
            raise ValueError('x_positive must be either "right" or "left"')


        ymin, ymax = self.ax.get_ylim()
        if ((ymin > ymax and self.y_positive == 'up') or
                (ymax > ymin and self.y_positive == 'down')):
            self.ax.set_ylim(ymax, ymin)
        elif ((ymax >= ymin and self.y_positive == 'up') or
                (ymin >= ymax and self.y_positive == 'down')):
            self.ax.set_ylim(ymin, ymax)
        else:
            raise ValueError('y_positive must be either "up" or "down"')
```

will have to be adapted lightly to work on `AxesSpec` rather than manipulating `matplotlib.axes.Axes` objects directly.

Finally, `LiveGrid` has access to the 'start' document and consults that to figure out whether it should "snake" the tiles of the grid:

```py
self.snaking = doc.get('snaking', (False, False))
...
if self.snaking[1] and (pos[0] % 2):
    pos[1] = self.raster_shape[1] - pos[1] - 1
```

First, notice that we only consult `snaking[1]`. By definition the first, slowest axis traversed by a grid scan cannot be "snaked" because it is only traversed once. That is `snaked[0]` is always `False`. Only the later axes, which are traversed repeatedly, can be snaked. Second, `Grid` only applies to 2D scans. So the only question we need to answer is, "Is the second axis snaked?" We can answer in the same way that `LiveGrid` does, by checking `run.metadata["start"]` to get the `doc` used in the snippet above.

Checklist for parity with [`LiveGrid` in bluesky](https://github.com/bluesky/bluesky/blob/888716e078dcc0f0479c0583e7a1a940af3f2524/bluesky/callbacks/mpl_plotting.py#L382):

- [ ] Accept `clim`, `cmap`, `extent`, `aspect` and pass through to `Image(..., style=...)`.
- [ ] Add support for manually setting x and y limits of `AxesSpec` and handle that in the `_matplotlib_axes.py` view code.
- [ ] Add `x_positive` and `y_positive` parameters and adapt code that handles them from `LiveGrid`.
- [ ] Handle "snaking".

danielballan:
@jklynch suggested that `RasteredImage` may be a better name for this. I'm good with that.

bluesky_widgets/models/plot_builders.py
```diff
@@ -806,8 +818,47 @@ def label_maker(run, field):
...
         self.figure = figure
 
+    @property
+    def cmap(self):
+        return self._cmap
+
+    @cmap.setter
+    def cmap(self, value):
+        for i in self.axes.images:
+            i.style.update({'cmap': value})
```
danielballan, line 834: This should update `self._cmap` as well, right?

bluesky_widgets/models/plot_builders.py
```diff
@@ -806,8 +818,47 @@ def label_maker(run, field):
...
+            i.style.update({'clim': value})
+
+    @property
+    def extent(self):
+        return self._extent
+
+    @extent.setter
+    def extent(self, value):
+        for i in self.axes.images:
+            i.style.update({'extent': value})
```
danielballan, line 852: Ditto.

bluesky_widgets/models/plot_builders.py
```diff
@@ -777,6 +785,10 @@ def __init__(
         needs_streams=("primary",),
         namespace=None,
         axes=None,
+        clim=None,
+        cmap='viridis',
+        extent=None,
+        aspect='equal',
```
danielballan, line 791: After talking through the various trade-offs yesterday, I came away with the feeling that we should remove this parameter entirely (along with the associated `property`) and leave it to the user to do this via either preconstructing `AxesSpec` manually or modifying it afterward directly.

```py
# Manually create Axes so as to set custom aspect.
axes = AxesSpec(aspect="...")  # might also set custom title or whatever they like...
rastered_image = RasteredImage(..., axes=axes)

# Or accept the default Axes but modify them after creation.
rastered_image = RasteredImage(...)
rastered_image.axes.aspect = "..."
```

My reasoning is that if we give special access to `aspect` here, why not also `title` and `x_label` and `y_label`.... it gets quickly messy.

In the block below in `__init__` where we auto-create at `AxesSpec` if one is not specifically given to us, we should be opinionated and pass in `aspect="equal"` since that is a more suitable default for `RasteredImage`.

danielballan (commented):
Seems on track. I left a couple suggestions.

bluesky_widgets/models/plot_builders.py
AbbyGi: Yep, I agree. I didn't make note of what we decided in my notes so I forgot about it

danielballan:
Looks good. Seems ready for the next items in the checklist.

bluesky_widgets/models/plot_builders.py
```diff
@@ -865,10 +865,15 @@ def _add_image(self):
...
         image = ImageSpec(func, self.run, label=self.field, style=style)
+        md = self.run.metadata["start"]
         self.axes.images.append(image)
         self.axes.title = self._label_maker(self.run, self.field)
-        self.axes.x_label = self.run.metadata["start"]["motors"][1]
-        self.axes.y_label = self.run.metadata["start"]["motors"][0]
+        self.axes.x_label = md["motors"][1]
+        self.axes.y_label = md["motors"][0]
+        if self.axes.x_limits is None:
+            self.axes.x_limits = (-0.5, md["shape"][1]-0.5)
```
danielballan, line 874: Would you add a comment to explain what this default behavior is doing, for the benefit of Future Us?

bluesky_widgets/_matplotlib_axes.py
```diff
@@ -137,8 +149,10 @@ def _add_image(self, image_spec):
         array = image_spec.func(run)
 
         # Initialize artist with currently-available data.
-        artist = self.axes.imshow(array, label=image_spec.label, **image_spec.style)
+        artist = self.axes.imshow(array, label=image_spec.label, **image_spec.style, origin='lower')
```
danielballan, line 152: Minor thing: let's put the unpacking parameter, `**image_spec.style`, last. Otherwise this leads me to wonder if there is something special going on with precedence (but there isn't) and "looks funny".

bluesky_widgets/_matplotlib_axes.py
```diff
@@ -149,10 +149,20 @@ def _add_image(self, image_spec):
...
         # Initialize artist with currently-available data.
-        artist = self.axes.imshow(array, label=image_spec.label, **image_spec.style, origin='lower')
+        artist = self.axes.imshow(array, label=image_spec.label, origin='lower', **image_spec.style)
         self.axes.set_aspect(self.model.aspect)
-        self.axes.set_xlim(self.model.x_limits)
-        self.axes.set_ylim(self.model.y_limits)
+        # Set x and y limits to defaults if they aren't defined
+        # This is needed for Image to have proper limits
+        if self.model.x_limits is None:
+            self.axes.set_xlim(-0.5, array.shape[1]-0.5)
```
danielballan, line 157: This steps outside the paradigm* that all changes slow from the model to the view. Can we make this change entirely on the model side when an image is first added to `AxesSpec`?


*The paradigm is important to avoid confusing race conditions where conflicting actions can drive the application into an inconsistent state. A good example of how this works is, when you close a tab in the view, it tells the model to remove the corresponding element, and the _model_ then tells the view to actually remove that tab. It would be more direct to just close the tab, but by telling the model first, we make sure that the model knows the tab is gone before we actually remove it.

danielballan:
Looks good. We should add some tests. Suggested by @jklynch:

- Use 2x2 grid for speed / simplicity.
- Test snaking.
- Test x_positive, y_positive.

bluesky_widgets/models/_tests/test_rastered_image.py
```diff
@@ -0,0 +1,85 @@
...
+    assert model.y_positive == 'up'
+
+
+def test_non_snaking_image_data(non_snaking_run):
+    run = non_snaking_run
+    model = RasteredImage("ccd", shape=(2, 2))
+    model.run = run
+    actual_data = model.figure.axes[0].images[0].func(run)
+    expected_data = [[1, 2], [3, 4]]
+    assert (actual_data == expected_data).all()
```
danielballan, line 76: The recommended solution here is [`numpy.array_equal`](https://numpy.org/doc/stable/reference/generated/numpy.array_equal.html). It provides more useful messages in the case of failure, and it provides a path for adding a _tolerance_ which is not relevant here but is often useful.

bluesky_widgets/models/_tests/test_rastered_image.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def test_snaking_image_data(snaking_run):
+    run = snaking_run
+    model = RasteredImage("ccd", shape=(2, 2))
+    model.run = run
+    actual_data = model.figure.axes[0].images[0].func(run)
+    expected_data = [[1, 2], [4, 3]]
+    assert (actual_data == expected_data).all()
```
danielballan, line 85: Ditto

danielballan (commented):
I left a minor implementation comment. I also think it would be useful to include a test or two that adds data incrementally so that can check that the snaking and non-snaking case display data in the correct pattern, not just that they arrive at the correct final result.

This could be done by using `RunBuilder` inside the test, and checking in on the state of things between calls to `add_data`, as in:

```py
with RunBuilder(md) as builder:
    ccd = [1, 2, 3, 4]
    x = [0, 1, 0, 1]
    y = [0, 1, 1, 0]
    # Initialize with just the first data point.
    builder.add_stream("primary", data={"ccd": next(ccd), "x": next(x), "y": next(y)})
   # Check on the figure.
    ...
    # Add another data point.
    builder.add_data("primary", data={"ccd": next(ccd), "x": next(x), "y": next(y)})
    # Check on the figure.
    # etc.
```

PR #76, opened 2020-12-10
Adding text search to search widget

bsobhani:
Search only works with the real mongo because mongoquery does not support text search.


danielballan:
Somehow a merge commit ended up here, which doesn't seem right. Doing some git surgery now (which is why the GitHub UI marked this as "closed"). Will post a summary of what I did when it's complete.

danielballan:
Please do not make PRs from your `master` branch. In general, collaborators will push to PR branches to provide help, and doing a PR from a `master` branch interferes with this. It's also better git hygiene to keep your `master` branch in sync with upstream, and do development work on feature branches.

I'll open a new PR with your commit.

danielballan:
See https://scikit-beam.github.io/scikit-beam/development/workflow/development_workflow.html#workflow

PR #77, opened 2020-12-10
adding search field

danielballan:
@bsobhani's commit from #76

### DIFF
--- a/bluesky_widgets/examples/qt_search.py
+++ b/bluesky_widgets/examples/qt_search.py
@@ -10,6 +10,34 @@ from bluesky_widgets.examples.utils.generate_msgpack_data import get_catalog
 from bluesky_widgets.examples.utils.add_search_mixin import columns
 from qtpy.QtWidgets import QWidget, QPushButton, QVBoxLayout
 
+def get_catalog2():
+    # generate_data.py
+    import logging
+    from suitcase.mongo_normalized import Serializer
+    from bluesky import RunEngine
+    from bluesky.plans import count
+    from ophyd.sim import det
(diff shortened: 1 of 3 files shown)


danielballan:
Added follow-up commits. See commit messages for details. The most important change is that `RunSearch` now checks to see whether its `catalog` supports `$text` queries and degrades gracefully if not. Given a msgpack-backed catalog, there is no text search UI:

image


and attempting to set `SeachInput.text` from the console raises an informative error:

```py
In [9]: app.searches.active.input.text = "something"
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-9-2cac80aa1b9d> in <module>
----> 1 app.searches.active.input.text = "something"

~/Repos/bnl/bluesky-widgets/bluesky_widgets/models/search.py in text(self, text)
    369     def text(self, text):
    370         if text and not self.text_search_supported:
--> 371             raise RuntimeError("This catalog does not support text search.")
    372         self._text = text
    373         self.events.text(text=text)

RuntimeError: This catalog does not support text search.
```

With this change to `bluesky_widgets.examples.qt_search`:

```diff
-from bluesky_widgets.examples.utils.generate_msgpack_data import get_catalog
+from bluesky_widgets.examples.utils.generate_mongo_data import get_catalog
```

a text search UI now appears

image

and can be updated from the console:

```py
In [4]: app.searches.active.input.text = "scan"
```

bsobhani:
Thanks, hiding the search if it is not supported is a good idea.

gwbischof (approved):

Issue #78, opened 2020-12-10
Put time entry inputs on one line to conserve vertical space.


danielballan:
As suggested by @cryos:

> Vertical space is always kind of precious in UIs [...] I wonder if the date might use "Date range: 'entry since' - 'entry until'" and only use one line where they are clearly related.

It currently has two lines, and it's clear that it would fit on one.

![image](https://user-images.githubusercontent.com/2279598/101799611-59defe80-3ada-11eb-9ff7-e49bf74a4562.png)

danielballan:
In that same vein, it looks like the "When:" radio buttons would fit just fine in an array of 2 rows vs 3 columns instead of the current 3 x 2 arrangement.

PR #82, opened 2020-12-18
Conserve vertical space

bsobhani:
Makes the 3x2 grid 2x3, and puts the since and until date selections on the same row.

### DIFF
--- a/bluesky_widgets/qt/_search_input.py
+++ b/bluesky_widgets/qt/_search_input.py
@@ -54,10 +54,10 @@ class QtSearchInput(QWidget):
         default_period_layout.setVerticalSpacing(10)
         default_period_layout.addWidget(self.all_widget, 0, 0, 1, 2)
         default_period_layout.addWidget(self.year_widget, 1, 0, 1, 2)
-        default_period_layout.addWidget(self.month_widget, 2, 0, 1, 2)
-        default_period_layout.addWidget(self.week_widget, 0, 1, 1, 2)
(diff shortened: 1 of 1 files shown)


danielballan:
Before (`master` branch):

image

After (this PR branch):

image

This looks like a good improvement to me.

Referencing the suggestion in #78:

> I wonder if the date might use "Date range: 'entry since' - 'entry until'"

I think @cryos had in mind

```
Date range: [since input] – [until input]
```

with an [en dash](https://en.wikipedia.org/wiki/Dash#En_dash) in between. This PR currently has it like:

```
Since: [since input] Until: [until input]
```

I think the original suggestion might be a little better, a little easier to for users to quickly grasp without thinking.

danielballan:
While we are here, should we change the label "24h" to "24 Hours"? I have internalized that abbreviation can be a barrier to non-native English speakers, and it seems we have plenty of space to spell it out here. It would also make it consistent with the other labels.

bsobhani:
I have no strong opinion either way but I made the changes so we can look at them:

date_range

danielballan:
That screenshot looks a bit cramped to my eye, but it's being shown in a very narrow window in that example. If I make the window just a bit wider so that there is no horizontal scrollbar anymore, the top area looks more reasonably spaced out.

image

I consulted @cryos briefly via Teams, and we are both happy to merge this as forward progress. Thanks, @bsobhani.

danielballan:
Confirmed locally that `flake8` and `pytest` pass, since Travis-CI is broken. (Needs migration to GH Actions.)

PR #87, opened 2021-01-12
Refactor AutoPlotter to be an abc. First examples of multi-axes figures.

danielballan:
This refactors `AutoLines` and `AutoImages` from a `functools.partial` to an abstract base class with one required method. This was developed in tandem with working on a use case for APS, sketched [in a gist](https://gist.github.com/AbbyGi/38b44e89ccd55f392be0dc9921820894).

## Design Features
* From the user perspective, `AutoLines` and `AutoImages` share the same user-facing API as normal, non-automatic plot builders like `Lines` and `Images`: they all have `add_run` and `discard_run`. That is easy to remember.
* That symmetry of API enables these to nest nicely. For example, an `AutoPlotAllTheThings` can fan out to `AutoLines`, `AutoImages`, and some custom plot builder (automatic or not) because they all implement `add_run` and `discard_run`.
* From a developer perspective, you just have to override one method, `handle_new_stream(self, run, stream_name)`, and manipulate two pieces of state from the base class, `self.plot_builders` and `self.figures`.

## Rejected Alternatives

* **Can we use composition instead of inheritance?** Decisions about whether and where to add or remove figures require access to and sometimes mutation of a lot of shared state. ("What figures do we have already? How many lines or images are they currently holding? What is the data (e.g. `BlueskyRun`s) corresponding to those?") This need for access to shared state, as well as the need to store application-specific state, drives us toward the mechanism of subclassing.
* **What about implementing the logic of what to plot as a function that returns a dict?** We briefly experimented with "plot suggestors" that return "plot suggestions", which one might describe as a "description of a description of a plot". This is more abstraction than we need at this time. There is a path to refactor to add that layer later if/when we need it and understand how to parameterize it correctly.

Developed through periodic discussions with @AbbyGi, @tacaswell, @dylanmcreynolds.


AbbyGi (approved):
Looks good to me.

PR #91, opened 2021-01-15
Switch to GH Actions. Factor plot model tests to run on installed views.

danielballan:


danielballan:
Self-merging to get GH Actions running

PR #94, opened 2021-01-15
Try using xvfb directly.

danielballan:
h/t @stuartcampbell


danielballan:
I like the directness of this approach. Install one package (which I notice GH already has installed in the base environment, so that's a no-op!) and set one env var. Avoid a complicated pytest integration.

stuartcampbell (approved):

Issue #65, opened 2020-12-04
Consider adding a layer to remove coupling with views and BlueskyRun API


danielballan:
The view code knows about of the concept of a `BlueskyRun` and its particular interface and signals:

https://github.com/bluesky/bluesky-widgets/blob/434046b54ee160cdcdd7a6c8c1fc1bf7e82d575b/bluesky_widgets/_matplotlib_axes.py#L104-L105

https://github.com/bluesky/bluesky-widgets/blob/434046b54ee160cdcdd7a6c8c1fc1bf7e82d575b/bluesky_widgets/_matplotlib_axes.py#L121-L125

What if, instead, we separate the bluesky-aware aspect into a separate components, perhaps by subclassing like so:

```py
class LineSpec:
    "Not bluesky-aware, just knows about an abstract `transform` (callable) and `data`
    def __init__(self, transform, data, ...):
        ...
        self.events = EmitterGroup(new_data=Event, completed=Event)

class BlueskyLineSpec(LineSpec):
    "Hooks up LineSpec's signals to BlueskyRun signals."
    def __init__(self, transform, run, ...):

        def forward_new_data(event):
            data = run[event.stream_name].to_dask()
            self.events.new_data(data=data)

        run.events.new_data.connect(forward_new_data)
        run.events.completed.connect(self.events.completed)
        ...
        super().__init__(transform, run, ...)

then `LineSpec` and the views become generic. That would extend the utility of these components at the cost of an additional layer of signaling.

danielballan:
Another important consequence of this is that the bottom layer `LineSpec` is not bound to a single `BlueskyRun`, and you could generate `LineSpec`s joining data from multiple Runs. And of course from things that have nothing to do with `BlueskyRun` at all.

danielballan:
Also, if `transform` always returns a `dict` of kwargs rather than an artist-specific construct (like `(x, y)` for lines) we can simplify the view code and reduce the artist-specific code.

danielballan:
#97 does this almost entirely. It moves the 1:1 correspondence to `BlueskyRun` into a classmethod. That could easily be moved off to a subclass in the future.

danielballan:
@tacaswell shared this snippet of work related to matplotlib 4.0 https://gist.github.com/tacaswell/95177903175dbc28be5353b4a0e5118f.  Keeping this here so I don't lose it.

PR #97, opened 2021-01-27
Refactor Model <---> View-model contract

danielballan:
This is a major refactor and API change of the lower-level entities. It does *not* break any high-level APIs. The user code that I know of will not need any changes to account for this.

## Description

* Bind `Line` and `Image` and in general our `Artist` model to an opaque callable that returns data instead of to a `BlueskyRun`.
* Reduce the API of `Axes`. Instead of keeping separate lists `axes.lines` and `axes.images`, as well as a view of their union, `artists`, just keep one heterogeneous list of `artists` and let the Views sort out what is what based on type. We were already relying on types in various places in the Views. It was not immediately clear to me that this "simplification" was a good one, but I am convinced by the result that it is.
* Rename `FigureSpec` -> `Figure`, `AxesSpec` -> `Axes`, etc. in response to considered feedback. See commit message for details.

## Motivation and Context

* A line should not have a 1:1 correspondence with a `BlueskyRun`. It may represent data aggregated from multiple `BlueskyRun`s. Or, perhaps in the future, from something that isn't a `BlueskyRun` at all.
* This simpler contract between Model and View-model makes it much easier to build new View-models to support new Views. We have near-term plans for a View that uses PIL, a proper napari View (going beyond the demo/example), and a Vega Lite one. This is a good time to refactor the contract to make that easier.

## How Has This Been Tested?

* More test parameterization to test models against all views

### DIFF
--- a/bluesky_widgets/_matplotlib_axes.py
+++ b/bluesky_widgets/_matplotlib_axes.py
@@ -1,16 +1,16 @@
 """
-This joins our AxesSpec model to matplotlib.axes.Axes. It is used by
+This joins our Axes model to matplotlib.axes.Axes. It is used by
 bluesky_widgets.qt.figures and bluesky_widgets.jupyter.figures.
 """
 import logging
 
-from .models.plot_specs import AxesSpec, LineSpec, ImageSpec
+from .models.plot_specs import Axes, Line, Image
 from .models.utils import run_is_live_and_not_completed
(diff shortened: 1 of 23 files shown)


danielballan:
This addresses #65 in large part but does not close it. It does as much of that as I think we _should_ do for now.

PR #99, opened 2021-01-27
Add an example for integration with an existing Qt app.

danielballan:
## Description

Our Qt examples so far fall into two categories:

1. Integration with specific complex Qt apps (napari, PyFAI)
2. Integration with IPython for interactive use

This example demonstrates integration with a dead-simple minimal Qt application, as a jumping off point for integration with Qt applications in general.

## Motivation and Context

For @JulReinhardt 

## How Has This Been Tested?

Run interactively.

```
python -m bluesky_widgets.examples.qt_app_integration
```

image

### DIFF
--- a/bluesky_widgets/examples/qt_app_integration.py
+++ b/bluesky_widgets/examples/qt_app_integration.py
@@ -1,5 +1,5 @@
 """
-An example integration QtFigures into an "existing" Qt applicaiton. Run like:
+An example integration QtFigures into an "existing" Qt application. Run like:
 
 python -m bluesky_widgets.examples.qt_app_integration
 """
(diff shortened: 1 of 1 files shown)


JulReinhardt:
Hi Dan, thanks for the example. I got a basic version running. Now I am trying to get it running using the queueserver. Where is the right place to define the figure model? Seems like I cannot do it in the qserver startup scripts. But it feels likes mixing realms when putting it into the GUI classes.

danielballan:
I agree. I'd suggest placing the code that configures/instantiates models in a module by itself, e.g.

```py
# plot_models.py

auto_lines = AutoLines(...)
# and probably additional, custom models as this grows...
```

and then importing those into wherever you are _instantiating_ your GUI classes (which may or may not be the same place that you _define_ them).

I'm sure we'll be wrestling with the best way to organize this too, and I'll let you know if we learn anything useful in the process.

PR #100, opened 2021-01-29
Avoid using pyplot

danielballan:
We currently abuse pyplot a little. This PR avoids using pyplot, but it also breaks interactive plotting in Jupyter. Until that is addressed, it shouldn't be merged.

A better solution might just be to avoid using matplotlib in Juptyer altogether.


danielballan:
The fundamental issue here, identified by @tacaswell, is that the kernel cannot process messages while executing a cell. This is the best workaround available without significant changes to ipykernel.

PR #102, opened 2021-02-01
Added field searching capabilities

### LINKED ISSUE #61: Make SearchInput easily extenible
Let's make it as easy as possible to add custom search fields to the form. This should be configured in the _model_ and then reflected in the view.

We might for example do:

```py
model = SearchInput(..., fields=['plan_name', 'proposal_id'])
```

which should add corresponding text input fields to the view. Do we want to think about making `fields` a dict so that we could provide `Enum`s that will be displayed as comboboxes in the view?

AbbyGi:
## Description

Add `fields` parameter to `SearchInput` class that adds text input fields to the view and allows for searching of specific fields.

## Motivation and Context

This makes it easier to add custom search fields to the search view.
Closes #61.

## How Has This Been Tested?

Tested interactively on example datasets, but still needs proper unit tests.

search_1

### DIFF
--- a/bluesky_widgets/models/search.py
+++ b/bluesky_widgets/models/search.py
@@ -7,6 +7,7 @@ import dateutil.tz
 
 from ..utils.list import EventedList
 from ..utils.event import EmitterGroup, Event
+from ..utils.dict_view import UpdateOnlyDict
 
 LOCAL_TIMEZONE = dateutil.tz.tzlocal()
 _epoch = datetime(1970, 1, 1, 0, 0, tzinfo=LOCAL_TIMEZONE)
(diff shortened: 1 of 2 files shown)


bluesky_widgets/models/search.py
```diff
@@ -270,9 +271,11 @@ def ensure_abs(*abs_or_rel_times):
 
 
 class SearchInput:
-    def __init__(self, *, text_search_supported=False):
+    def __init__(self, *, fields=None, text_search_supported=False):
         self._since = None
         self._until = None
+        self._fields = fields or []
+        self._field_search = UpdateOnlyDict({field: None for field in self._fields})
```
danielballan, line 278: Therre a handy classmethod on `dict` for constructing things like this

```suggestion
        self._field_search = UpdateOnlyDict({dict.fromkeys(self._fields))
```

danielballan (commented):
Looks great. Just one very minor implementation suggestion that you can take or leave.

It just needs a unit test or two, as you noted in the description. Great write-up!

PR #103, opened 2021-02-10
WIP: Add option for displaying figures below cells that create them.

### LINKED ISSUE #95: Use case: place a fresh canvas in the currently-executing Jupyter notebook cell.
At FMX, they measuring rocking curves in a Jupyter notebook and refer back to the notebook as a logbook of what was done. They _do not_ want any over-plotting. They want something like the user experience of `%matplotlib inline`, but with a live-updating plot during acquisition. It also wouldn't hurt to retain interactivity (i.e. pan and zoon) after the scan, but this is not essential.

To address this, would could have a `JupyterFigures` variant that places each new figure in the currently-executing cell (by calling `display(...)` rather than adding a tab to an exiting widget. I am not 100% sure this is technically feasible, so we'll need to try it.

danielballan:
## Description

It is often useful to place the figure(s) in a tabbed viewer in one cell, rather than "plopping" them into the output area below the currently-running cell. However, there is still a use case for this mode of placing figure "in line" with the work. (Not to be confused with `%matplotlib inline` which kills interactive pan and zoom.) See #95.

See screenshot below.

TO DO

- [x] Add option for placing figures below running cell.
- [ ] Embed figures so that they can be viewed if the notebook is opened later. The action item here is probably just to wait for ipympl. They are [working on this](https://github.com/matplotlib/ipympl/pull/294). Perhaps, while we wait, we could implement `%matplotlib inline`-like behavior as a stop-gap. That is, when cell execution completes, close the figure and replace it with an embedded PNG.
- [ ] Tests

## Motivation and Context

Closes #95. See that issue for motivation and context.

## How Has This Been Tested?

Needs tests

### DIFF
--- a/bluesky_widgets/jupyter/figures.py
+++ b/bluesky_widgets/jupyter/figures.py
@@ -1,4 +1,5 @@
 from ipywidgets import widgets
+from IPython.display import display
 import ipympl.backend_nbagg
 import matplotlib.figure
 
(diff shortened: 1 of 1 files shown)


mrakitin (approved):
Thanks for the update!

PR #104, opened 2021-02-10
Fix outstanding TODO about keeping up with documents.

danielballan:
## Description

This makes the 0MQ `RemoteDispatcher` more robust, though we have more still to do.

## Motivation and Context

Previously, each worker read one document out of the socket, returned it, and quit. Now, each worker reads document(s) from the socket, yields them one at a time for processing, and quits once the socket is empty. This means that if the documents are generated faster that the sleep interval between launching workers, we _might_ keep up. It does not yet handle the question of what happens if the documents build up faster than we can read them out and handle them off. But it is a meaningful step forward.

## How Has This Been Tested?

I ran the example

```
python -m bluesky_widgets.examples.advanced.qt_viewer_with_search localhost:XXXXX
```

to confirm that this at least does not make anything worse. More thorough testing is needed to verify that it can handle a fast document-generation load.

### DIFF
--- a/bluesky_widgets/qt/zmq_dispatcher.py
+++ b/bluesky_widgets/qt/zmq_dispatcher.py
@@ -65,22 +65,24 @@ class RemoteDispatcher(QObject):
 
     def _receive_data(self):
         our_prefix = self._prefix  # local var to save an attribute lookup
-        # TODO Pull on the socket more than once here, until it blocks, to
-        # ensure we do not get more and more behind over time.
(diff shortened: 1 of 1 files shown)


dmgav (approved):
Looks good to me. Would it make sense to cache the documents by placing them in the queue in case processing can not keep up with document flow. It is typical for processing time to be non-uniform. For example an application may accumulate certain number of data points and then process them as a batch. On the other hand, caching may be delegated to application developer.

PR #106, opened 2021-02-17
ENH: auto-live-table

tacaswell:
This is still work done in prep of implementing an auto LiveTable.

Unlike plotting where all-of-dataset access is helpful, tables are well -suited to streaming ("print the latest news"), so this is likely going to depend on https://github.com/bluesky/bluesky-live/pull/15 going in so we can mostly re-use the logic from LiveTable.


danielballan:
The ipywidgets ecosystem finally has a performant table widget, just in time for us. https://github.com/bloomberg/ipydatagrid/blob/main/examples/DataGrid.ipynb

tacaswell:
Rebased.

bluesky_widgets/models/auto_plot_builders/_base.py
```diff
@@ -73,6 +73,14 @@ def handle_figure_removed(self, figure):
                 if figure is plot_builder.figure:
                     self.plot_builders.remove(plot_builder)
             elif hasattr(plot_builder, "figures"):
+                # This code looks broken, was untested, and TAC is not
```
danielballan, line 76: I can't figure out what looks broken about this. I think it works and is being used.

PR #107, opened 2021-02-23
Add example with RE manager.

danielballan:
## Description

This adds an example with the following data flow:

Untitled Diagram

## Motivation and Context

First step toward a RunEngine GUI

## How Has This Been Tested?

Tested interactively.

image

### DIFF
--- a/bluesky_widgets/examples/advanced/qt_with_RE_worker.py
+++ b/bluesky_widgets/examples/advanced/qt_with_RE_worker.py
@@ -67,7 +67,7 @@ def main():
                 time.sleep(0.5)
 
         # Add plan to queue
-        client.send_message(
+        response = client.send_message(
             method="queue_item_add",
             params={
                 "plan": {"name": "scan", "args": [["det"], "motor", -5, 5, 11]},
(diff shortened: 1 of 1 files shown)


dmgav:
I think the diagram in the description is inaccurate. I would rename `Run Engine` box into `RunEngine worker` (similar to `RunEngine manager` box). The documents are not returned to RunEngine manager, they are emitted by RunEngine worker via 0MQ proxy. So the arrow `Bluesky documents` should go directly from RunEngine worker to 0MQ proxy.

danielballan:
I was 50/50 on how to do that. The documents do not flow through the manager, but the manager is in charge of controlling where they go because it sets up the subscription. That is, the `localhost:5577` config goes to the manager, not directly to the RE. Maybe this diagram just isn't the right way to show it.

dmgav:
The role of the manager is to configure and start RE. But I am afraid that if we add things, the diagram may become difficult to understand.

dmgav (approved):

dmgav:
May be we can put two arrows going from RunEngine manager box to Run Engine box: one is 'plans', the other is '0MQ data config' (or something similar) and then show that documents are forwarded to 0MQ proxy directly.

PR #111, opened 2021-03-09
Add example of integrating search into app.

danielballan:
## Description

Example of integrating the Qt search widget, plus a button, into an "existing" application.

1. Edit `CATALOG_NAME` to point to one of the names in `list(databroker.catalog)` on your system (e.g. `"iss"`).
2. Start it from a terminal like

    ```
    python -m bluesky_widgets.examples.qt_app_search_integration
    ```

3. Select some runs and click the "Open" button at the bottom.
4. Check to see if a corresponding list of `BlueskyRun`s was printed in the terminal.

## Motivation and Context

For ISS

image


AbbyGi (approved):
Looks good. Merging.

PR #112, opened 2021-03-10
Make ys mutable in Lines

### LINKED ISSUE #69: Support adding and removing ys from RecentLines
The task here is to make `RecentLines.ys` an observable list instead of an immutable tuple. When its contents are changed, lines must be correspondingly added or removed. Also, the axes y label and title should be updated.

For now I propose to _not_ make `x` settable. To change `x`, just make another plot. Three reasons:

1. If you change `x` everything has to re-plotted anyway, so it isn't much more expensive to just create a new `RecentLines` instance from scratch.
2. In a GUI, we don't need to expose this asymmetry to the user. We can make changing x look like just changing y; they don't need to know that we'll be making a new `RecentLines` instance internally when they change x.
3. It seems to be useful for higher layers (e.g. `AutoRecentLines`) to treat `x` as fixed, so I'd like to leave it that way unless we hit a good reason to make it settable.

AbbyGi:
## Description

This makes the `ys` in `Lines` an observable list instead of an immutable tuple. Lines can now be added and removed from the plot. `Lines` will also update the title and y label to keep track of existing `ys` if the `Axes` doesn't specify the title or y label. Users can update the title or y label and `Lines` will not change them unless they are set to `None`.

## Motivation and Context

This will allow lines to be added and removed in `Lines` without having to make a new instance of `Lines`. Closes #69.

## How Has This Been Tested?

Tested interactively by running the `ipy_qt_figures.py` example and modifying `model.ys`, `model.title`, and `model.y_label`. Unit tests are included.

bsw_pr_112_1

bsw_pr_112_2


bluesky_widgets/models/plot_builders.py
```diff
@@ -203,6 +207,13 @@ def set_figure(event):
         self.add_run = self._run_manager.add_run
         self.discard_run = self._run_manager.discard_run
 
+        self.ys.events.added.connect(self._add_ys)
+        self.ys.events.removed.connect(self._remove_ys)
+
+    def _map_ys_to_artists(self, artist, y):
```
danielballan, line 213: Does this method buy us anything? Can the callers just touch `self._ys_to_artist` directly? If we anticipate that more logic will be added here in time, that would be a good reason to keep it, but I don't see that yet from a quick read.

bluesky_widgets/models/plot_builders.py
```diff
@@ -233,6 +245,49 @@ def restyle_line_when_complete(event):
...
             self.axes.artists.append(line)
+            self._ys_to_artists[y].append(line)
+
+    def _add_ys(self, event):
+        # Update title and y_label when adding a new y
+        self.y_label = ", ".join(auto_label(y) for y in self.ys)
+        self.title = f"{self.axes.y_label} v {self.axes.x_label}"
+        y = event.item
+        if y in self._ys_to_artists.keys():
+            # Don't plot a new line if y already exists
```
danielballan, line 256: I'm worried this could create unexpected behavior if someone _wants_ to plot the same thing twice or some other unforeseen scenario. It just feels odd to have the let `self.ys` get out of sync with `self._ys_to_artists`. If we think it's important to protect users from plotting something twice, I'd suggest we should block them from doing so further up---i.e. reject the addition to `self.ys` in the first place by giving it the semantics of an `OrderedSet` rather than a list. (That might be something to consider in a separate issue/PR.)

danielballan (commented):
Looks great. One point noted in line needs some consideration, along with the stuff we're discussing concurrently on Slack about Axes title, but this looks close to ready.

bluesky_widgets/models/plot_builders.py
```diff
@@ -196,18 +195,48 @@ def set_figure(event):
 
             self.axes.events.figure.connect(set_figure)
 
+        # Keep title up to date with self.ys or leave it as user-defined value
+        if self.axes.title is None:
```
danielballan, line 199: Minor thing: this can be consolidated to:

```py
self._control_title = self.axes.title is None
```

and similar.

bluesky_widgets/models/plot_builders.py
```diff
@@ -196,18 +195,48 @@ def set_figure(event):
...
+            self._control_title = True
+        else:
+            self._control_title = False
+        # Keep y_label up to date with self.ys or leave it as user-defined value
+        if self.axes.y_label is None or self.axes.y_label == self._default_y_label():
```
danielballan, line 204: And similarly here.

danielballan (commented):
Minor implementation suggestion. Otherwise this looks good to go, though I'd like to take it for a spin interactively before merigng.

danielballan:
Next step here is to look into why CI is failing.

danielballan:
The tests look good. I tested interactively as well, for good measure. This is really fun to use!

Issue #64, opened 2020-12-03
Specify the optional dependencies.


danielballan:
The only hard dependency we have is `bluesky-live` (where the models' signaling system is defined). Most components have additional particular dependencies. The views have dependencies by framework. The models have dependencies by function. It should be possible to just install the ones you need, as in:

```
pip install bluesky-widgets[PyQt5,search]
pip install bluesky-widgets[jupyter,plots]
```

and an easy kitchen-sink option if you don't mind pulling in a bunch of dependencies

```
pip install bluesky-widgets[complete]
```

danielballan:
This pattern may be useful here.

https://github.com/bluesky/catalog-server-from-scratch/blob/main/setup.py

PR #113, opened 2021-03-10
Read optional dependencies from files

jklynch:
Add a requirement file for each named installation "extra" and special processing for those files in `setup.py`.

Addresses #64.


jklynch:
Would this be a good PR to add installation instructions to README?

danielballan:
Thanks for this.

PR #114, opened 2021-03-10
Start work on models for RE client

danielballan:
This is not at all ready for review or comment, just a start.

### DIFF
--- a/bluesky_widgets/models/run_engine_client.py
+++ b/bluesky_widgets/models/run_engine_client.py
@@ -2,7 +2,7 @@ import time
 
 from bluesky_live.list import ListModel
 from bluesky_live.event import EmitterGroup, Event
-from bluesky_queueserver.manager.comms import ZMQCommSendThreads
+from bluesky_queueserver.manager.comms import ZMQCommSendThreads, CommTimeoutError
 
 
 class PlanItem:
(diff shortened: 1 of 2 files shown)


danielballan:
FYI @cryos, @dmgav has begun looking at this as well and may be adding commits to this branch.

bluesky_widgets/models/run_engine_client.py
```diff
@@ -0,0 +1,372 @@
...
+
+class PlanQueue(ListModel):
+    pass
+
+
+class PlanHistory(ListModel):
+    pass
+
+
+class BigModel:
```
danielballan, line 50: Is this needed? It looks like it isn't used it the demo. Maybe we shouldn't add this until/unless need it (especially since it was evidently hard to think of a good name).

bluesky_widgets/models/run_engine_client.py
```diff
@@ -0,0 +1,372 @@
...
+                    method="status", raise_exceptions=True
+                )
+                self._re_manager_accessible = True
+            except CommTimeoutError:
+                self._re_manager_accessible = False
+            if (status != self._re_manager_status) or (
+                accessible != self._re_manager_accessible
+            ):
+                # Status changed. Initiate the updates
+                self.events.status_changed()
```
danielballan, line 113:
```suggestion
                self.events.status_changed(status=self._re_manager_status)
```

bluesky_widgets/models/run_engine_client.py
```diff
@@ -0,0 +1,372 @@
...
+
+    def clear(self):
+        # Clear the queue.
+        response = self._client.send_message(method="queue_clear")
+        if not response["success"]:
+            raise RuntimeError(f"Failed to clear the plan queue: {response['msg']}")
+
+    def clear_online_status(self):
+        self._re_manager_accessible = None
+        self.events.status_changed()
```
danielballan, line 93: It might be convenient for the callbacks to receive the `status` in the `Event`. (Any keyword arguments are allowed here, and then become attributes on the emitted `Event`.)

```suggestion
        self._re_manager_status.clear()
        self.events.status_changed(status=self._re_manager_status)
```

bluesky_widgets/models/run_engine_client.py
```diff
@@ -0,0 +1,372 @@
...
+
+    def load_re_manager_status(self, *, enforce=False):
+        if enforce or (
+            time.time() - self._re_manager_status_time
+            > self._re_manager_status_update_period
+        ):
+            status = self._re_manager_status
+            accessible = self._re_manager_accessible
+            try:
+                self._re_manager_status = self._client.send_message(
```
danielballan, line 103: Could we use `self._re_manager_status.update(...)` here so that another part of the code holding a reference to `self._re_manager_status` will not end up with a stale reference when the content changes?

If response does not always includes the same keys, then we'll need `clear()` followed by `update()`.

bluesky_widgets/models/run_engine_client.py
dmgav: I didn't delete any of the preexisting code yet. But `BigModel` could probably be deleted. Regarding the new code, I think there should be some renaming done once the first working prototype is ready.

bluesky_widgets/models/run_engine_client.py
dmgav: I am careful not to hold many references, but it may be a good idea to change it to prevent future problems.

dmgav:
@danielballan I implemented all the suggestions. I guess you can merge this PR to keep `bluesky-widget-demo` working. I will create a different branch for further commits.

danielballan:
Thanks, @dmgav. That sounds good.

PR #119, opened 2021-03-18
add support for Kafka

jklynch:
This PR adds a RemoteDispatcher for Kafka messages. It is modeled on the 0MQ RemoteDispatcher.


jklynch:
Trying to get kafka working in a GitHub workflow sent me down a rabbit hole for a long time. In the end I tightened up the bluesky-kafka tests and added a couple minor features, but still can't get the tests to pass reliably even though they pass on my laptop and were pretty well-behaved on travis. I may leave out a full kafka test in bluesky-widgets.

jklynch:
@danielballan the missing files are in place now but the test is not passing yet.

PR #121, opened 2021-04-02
Improvement of visual representation of the table representing catalog search results

dmgav:
## Description
Insignificant changed to the code that improve visual representation for catalog search results:

- Changed column name `Transient Scan ID` to `Scan ID`: the cell value is unlikely to be more than 5 characters wide and is 1 or 2 characters wide in the example. The table looks much better if the column is narrower. There are 2 more instances of `Transient Scan ID` in the code, which could also be replaced by `Scan ID` for consistency if needed.

- Change table settings to show grid, alternate row colors, make the last column extendable and column header titles left-aligned.

-  Use `QSpacerItem` to fill the space below the catalog name selector (`QComboBox`) when no catalog is selected.

- Deselect catalog (clear catalog name from the combo box) after 'Back' button is pressed.


AbbyGi (approved):

danielballan:
The linting failed because `black` accidentally removed their `stable` branch, which broke our linting GitHub Action workflow (and many others). Looks like they plan to restore it, so we don't need to change anything about our workflow. https://github.com/psf/black/issues/2079

Confirmed locally that `black` runs cleanly on this branch.

mrakitin:
@danielballan, is the local `black` from a release on PyPI? Would it make sense to pin to a specific version corresponding to that latest release? I understand there may be more work in the future to make sure we use the latest features, but in most cases, the local tests use the already installed version which also may be out-of-date.

Alternatively, [black.yml](https://github.com/bluesky/bluesky-widgets/blob/master/.github/workflows/black.yml) may be reworked to look like [flake8.yml](https://github.com/bluesky/bluesky-widgets/blob/master/.github/workflows/flake8.yml). That way we would rely on the more "stable"/static releases on PyPI.

dmgav:
I think that it makes a lot of sense to rework `black.yml` to look like `flake8.yml` instead of relying on GH actions. Let me try how well it works.

dmgav:
I implemented changes in PR https://github.com/bluesky/bluesky-queueserver/pull/147. The modified action seems to work fine: 'black' is downloaded from PyPI, action fails if 'black' is expected to change any of the files and succeeds otherwise.

mrakitin:
I like that. Thanks, @dmgav!

danielballan:
I agree.

PR #125, opened 2021-04-15
Expose which figure is active

AbbyGi:
## Description

Added a new `active` property to `Figure` and a new `active_index` property to `FigureList`. `QtFigures` also updates the active figure when swapping between tabs.

## Motivation and Context

We need to be able to access the current figure in order to add plot builders to the correct plot.

## How Has This Been Tested?

Tested interactively with the `bluesky-widgets-demo`.

### DIFF
--- a/bluesky_widgets/models/plot_specs.py
+++ b/bluesky_widgets/models/plot_specs.py
@@ -544,7 +544,7 @@ class Image(ArtistSpec):
 
 
 class FigureList(EventedList):
-    __slots__ = ("_active_index")
+    __slots__ = ("_active_index",)
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._active_index = None
--- a/bluesky_widgets/models/plot_specs.py
+++ b/bluesky_widgets/models/plot_specs.py
@@ -45,7 +45,7 @@ class Figure(BaseSpec):
         should fall back on ``title`` if this is None.
     """
 
-    __slots__ = ("_axes", "_title", "_short_title", "_active")
+    __slots__ = ("_axes", "_title", "_short_title")
 
     def __init__(self, axes, *, title, uuid=None, short_title=None):
         for ax in axes:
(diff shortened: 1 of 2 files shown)


danielballan:
I think that the state of which is active should live on the container (`FigureList`) rather than on the item (`Figure`). Some reasons for this:

* A` Figure` that isn't part of a `FigureList` shouldn't have vestigial `active` attribute that doesn't really mean anything.
* It's possible that a `Figure` could be an item in two or more `FigureList`s, where it is "active" in one but not in another. This implementation gives no way to represent that state.

My own implementation of [`SearchList`](https://github.com/bluesky/bluesky-widgets/blob/7dabea7e656374e90acc3981cf1e41c887a8b691/bluesky_widgets/models/search.py#L613), which was based on following napari's `LayerList`, took this same approach, which it retrospect I think is flawed. (The approach is more defensible in napari's case, but it does not apply as well to ours.) I don't think we should spread it to `FigureList`; instead, we should go back and fix it.

Could you refactor along the lines of:

```py
class FigureList(...):
    def __init__(...):
        self._active_index = None
        ...
```

where the `FigureList` knows which of its items is active, but the `Figure` is unaware where it is "active" or not in any given list?

bluesky_widgets/models/plot_specs.py
```diff
@@ -544,7 +544,21 @@ class Image(ArtistSpec):
 
 
 class FigureList(EventedList):
-    __slots__ = ()
+    __slots__ = ("_active_index")
```
danielballan, line 547:
```suggestion
    __slots__ = ("_active_index",)
```

bluesky_widgets/qt/figures.py
```diff
@@ -155,6 +157,10 @@ def _on_title_changed(self, event):
             index = self.indexOf(widget)
             self.setTabText(index, event.value)
 
+    def _on_tab_changed(self, event):
```
danielballan, line 160: This method is being used as a call by a Qt Signal (that is, it's a Qt Slot) not a call back for bluesky-live EventEmitters, so I think we should name this parameter `index` or something other than `event`.

danielballan (commented):
This looks good assuming it works. One tiny but important fix suggested in line.

danielballan (approved):

PR #126, opened 2021-04-16
Handle initalized array.

danielballan:
Bug-fix
## Motivation and Context

Observed while prepping https://github.com/bluesky/tutorial-adaptive-RL-sampling

## How Has This Been Tested?

Needs test


danielballan:
I think this bug was observed at PDF.

AbbyGi (approved):
Looks good after talking it through.

PR #129, opened 2021-04-27
Jupyter SearchInput

AbbyGi:
## Description

Added a Jupyter view for the SearchInput model.

## Motivation and Context

Jupyter doesn't have any search views yet. We need to make sure Jupyter has the same capabilities as Qt does.

## How Has This Been Tested?

Tested interactively (see screenshot below). Tested changing model properties and changing GUI values.

jupyter_search_input

### DIFF
--- a/bluesky_widgets/jupyter/_search_input.py
+++ b/bluesky_widgets/jupyter/_search_input.py
@@ -173,8 +173,10 @@ class JupyterSearchInput(HBox):
 def _blocked(widget, handler, names=All, type="change"):
     "Block signals from this object inside the context."
     widget.unobserve(handler, names, type)
-    yield
-    widget.observe(handler, names, type)
+    try:
+        yield
+    finally:
+        widget.observe(handler, names, type)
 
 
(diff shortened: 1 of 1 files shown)


bluesky_widgets/jupyter/_search_input.py
```diff
@@ -0,0 +1,180 @@
...
+    def _now():
+        return datetime.now(tz=LOCAL_TIMEZONE).replace(second=0, microsecond=0)
+
+
+@contextlib.contextmanager
+def _blocked(widget, handler, names=All, type="change"):
+    "Block signals from this object inside the context."
+    widget.unobserve(handler, names, type)
+    yield
+    widget.observe(handler, names, type)
```
danielballan, line 177: We should ensure that the `handler` is restored (i.e.`observe(..)` is run) even if an error is raised by the code run during the `yield` stage. See https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager for example.

danielballan (changes requested):
Looks good. One important change to make then, this should be ready to go in.

PR #131, opened 2021-05-04
ENH: generate 'history_item_process' event

dmgav:
Generate `history_item_process` event of the model `RunEngineClient` in response to double-clicking of an item in the history table. The event is delivering the dictionary of the selected history item parameters as a parameter `item`. The exit status of the plan may be found as `item["result"]["exit_status"]` and the list of run UIDS as `item["result"]["run_uids"]`.

It is recommended to verify that the keys "result", "exit_status" and "run_uids" are present in the dictionary to prevent crashes in case the history contains corrupted items.


danielballan (approved):

danielballan:
Good to go once @AbbyGi takes a look.

AbbyGi (approved):

PR #136, opened 2021-05-18
Plan Editor: Option to upload plans from a spreadsheet

dmgav:
Extention of a Plan Editor Widget: implementation of an option that allows uploading plans from a spreadsheet. Plan Editor widget is now contains an additional button 'Batch Upload':
image

Clicking on the `Batch Upload` button opens `Batch Upload` dialog box:
image

The dialog box contains the button `..`, which opens standard dialog for selection of an existing file and optional combo box for selecting spreadsheet type. The list of spreadsheet types (strings) may be set during initialization of the model. If the list is not set, then the combo box is not displayed:
image

Once the file is selected, the `Ok` button is enabled.
image

Pressing the `Ok` button adds the plans to the queue. The block of newly inserted items is automatically selected so that queue operations (copy/delete/move) could be applied to the block.
image

### DIFF
--- a/bluesky_widgets/models/run_engine_client.py
+++ b/bluesky_widgets/models/run_engine_client.py
@@ -1,11 +1,14 @@
 import collections
 import copy
+import os.path
 import pprint
 import time
+import importlib
 
 from bluesky_live.event import EmitterGroup, Event
 from bluesky_queueserver.manager.comms import ZMQCommSendThreads, CommTimeoutError
 from bluesky_queueserver.manager.profile_ops import bind_plan_arguments
+from bluesky_queueserver.manager.conversions import spreadsheet_to_plan_list
 
 
 class RunEngineClient:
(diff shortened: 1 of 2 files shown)


bluesky_widgets/models/run_engine_client.py
```diff
@@ -58,6 +61,13 @@ def __init__(self, zmq_server_address=None, user_name="GUI Client", user_group="
         #   using positions is consistent. Empty list - no items are selected
         self._selected_history_item_pos = []
 
+        # Current directory (used for opening/saving files).
+        self.current_dir = None
```
danielballan, line 65: Should this set to `os.cwd()` by default?

bluesky_widgets/models/run_engine_client.py
```diff
@@ -58,6 +61,13 @@ def __init__(self, zmq_server_address=None, user_name="GUI Client", user_group="
...
+        # Set the variable to the name of the Queue Server Custom Module (if available)
+        self.qserver_custom_module_name = None
```
danielballan, line 67: Should this be a list of modules?

dmgav: No, just the name of the module or None if there is no such module. I am using environment variable to pass the module name to HTTP server.

bluesky_widgets/models/run_engine_client.py
dmgav: It is set to `os.cwd()` later. I think general flow of data and setting should be revisited at certain point to take into account the most of the possible use cases.

PR #138, opened 2021-05-18
updates for bmm demo

gwbischof:


danielballan:
This got us through the demo, but I don't think it should be merged.

The general problem here is that the figure title may contain any string, but not all strings are valid filenames. A solution like https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote_plus that will cover all the cases is needed.

Issue #141, opened 2021-05-24
RE Widgets: horizontal scrolling issue in Queue Widget


dmgav:
If the plan parameters are represented as one long string (e.g. parameters are represented as `kwargs` argument), the 3rd column of the Queue Widget table is very wide. Clicking on an item in the 3rd column activates the item and makes the table automatically scroll to the right to display most of the activated item in column 3, making columns 1 and 2 invisible until user manually scrolls the table to the left. This behavior is not expected occurs if the plans are properly annotated and parameters are represented as short string, but it is may be annoying to the user when it happens. 

## Expected Behavior
The table should not automatically adjust horizontal scrolling to display active item. Horizontal scrolling change only if the user manually manipulates the scroll bar.

dmgav:
Addressed in PR https://github.com/bluesky/bluesky-widgets/pull/143

PR #143, opened 2021-06-06
Fix for the horizontal scrolling issue in queue and history tables

dmgav:
The changes include:
- Disabled autoscrolling (the table is not automatically scrolled to fully display the active item).
- Save and restore horizontal scrolling position while selection is changed programmatically.

As a result of the changes, the horizontal scrolling is consistently maintained as it is set manually by
the user when the items are added and removed from the queue/history or items are selected and 
deselected.

Addresses issue https://github.com/bluesky/bluesky-widgets/issues/141


danielballan (approved):

AbbyGi (approved):
New changes look good. Can be merged after being rebased.

Issue #140, opened 2021-05-24
RE Widgets: automatically switch plan editor to 'Edit' mode for a new plan


dmgav:
Automatically switch Plan Editor to Edit mode for a new plan once the user starts editing plan parameters. The proposed changes will minimize the number of actions required from the user to add items to the queue and will make the default plan editing interface more intuitive.

## Expected Behavior
Currently the Plan Editor displays parameters for the plan selected in the combo box. The parameters are read-only. The editor is switched into 'edit' mode by pressing the `New` button. Pressing the `New` button adds an additional step to the process of adding each new plan. 

Proposed changes to the behavior of the plan editor:
- Once a plan is selected in the combo box, the parameters are displayed in the table. Selecting different plan will change the contents of the table.
- The parameters of the selected plan are editable. The first attempt to modify a parameter will switch the editor into 'edit' mode, i.e. the plan selection combo box becomes disabled until editing is completed (plan is added or editing is cancelled).
- Once all the parameters are set to acceptable values, the option (push buttons) to add the plan to the queue becomes enabled.
- If the plan/instruction has no parameters or all parameters are optional, the option of adding the plan to the queue will be enabled once the plan is selected from the combo box, i.e. those plans/instructions could be added to the queue without switching into 'edit' mode.
- Most likely the 'New' button will be removed.

dmgav:
Addressed in https://github.com/bluesky/bluesky-widgets/pull/145

PR #144, opened 2021-06-07
Usability of plan editor: automatically switch to 'edit' mode once editing starts

dmgav:
Implements changes that improve usability of the plan editor by reducing the number of clicks that are necessary to add a new plan or instruction to the queue.

## Description

In the initial implementation of the plan editor, adding a new plan or instruction required the user to perform the following actions:
- Select the name of the plan or instruction in the Combo Box. This would open the plan form (in the form of a table) in Preview Mode, which allows to view plan parameters. Plan parameters are not editable in this mode.
- Click 'New' Button. This would switch the editor to the Edit Mode and make parameters editable. This would also enable 'Cancel' and 'Reset' buttons.
- Type the values of the item parameters in the table. Once validation of parameters is successful, 'Add To Queue' button is activated.
- Edit Mode could be exited by clicking 'Add To Queue' button (adds the item to queue) or pressing 'Cancel' button (discards changes). The form can be cleared (returned to the initial state) by pressing 'Reset' button.
- Opening another plan for editing (e.g. by pressing 'Edit' button in the plan viewer) cancels editing of the current plan, discards changes and loads the other plan for editing. Editing of plan parameters is expected to be a trivial operation, so asking for confirmation each time another plan is opened for editing will just annoy the user.

Changes in this PR:
- 'New' button is removed.
- Plan parameters are editable in the Preview Mode.
- The editor is automatically switched to Edit Mode after the first attempt to edit plan parameters.
- Plan parameter validation is enabled in preview mode. If a plan or instruction has no parameters or all parameters are optional, i.e. they have default values, the 'Add To Queue' button is enabled in the Preview Mode and the item can be added to the queue without switching to the Edit Mode.

The updated sequence of actions required from the user to add a plan to the queue:
- Select the name of the plan or instruction in the Combo Box. This would open the plan form ( table) in Preview Mode. 
- If plan has not parameters or all the parameters are optional (have default values), and no parameter values need to be changed then 'Add To Queue' button is enabled and the item can be added to the queue right away.
- User may start editing plan parameters in the Preview Mode. The editor will automatically switch to Edit mode once one of the parameters is changed. 
- Once the editor is switched to Edit Mode: radio buttons and combo box for selection of item type and name are disabled, 'Cancel' and 'Reset' buttons are enabled. 'Reset' button may be used at any time to clear the form (reset it to the original state, which is an empty form for the new plan), and 'Cancel' button may be used to discard editing results and switch back to the Preview Mode.
- Once the correct values of the parameters are entered into the form, user may press 'Add To Queue' button. This will add the item to the queue, switch the editor in the Preview Mode and activate Plan Viewer tab displaying recently added item.
- The added item may be opened for editing by pressing 'Edit' button in the Plan Viewer tab.
- If another plan is opened for editing while a plan is open in the Plan Editor (e.g. by pressing 'Edit' button in the Plan Viewer tab), editing of the current plan is canceled, changes to parameters are discarded and the other plan is loaded. Editing of plan parameters is expected to be a trivial operation, so asking for confirmation each time the parameters are discarded while another plan is opened for editing will simply annoy the users without any benifits.

Addresses the issue https://github.com/bluesky/bluesky-widgets/issues/140

## Motivation and Context
While preparation for the demo it was observed that the Plan Editor requires the user to perform a series of awkward operations. The goal of this PR is to simplify the process of adding new plans to the queue and make it more intuitive. After the proposed changes it is reasonable to assume that a new user could add the plan to the queue without training or too much experimentation.

### DIFF
--- a/bluesky_widgets/qt/run_engine_client.py
+++ b/bluesky_widgets/qt/run_engine_client.py
@@ -2081,7 +2081,6 @@ class _QtReEditor(QWidget):
         # self._combo_item_list.setSizePolicy(QComboBox.AdjustToContents)
         self._combo_item_list.currentIndexChanged.connect(self._combo_item_list_sel_changed)
 
-        self._pb_new_item = QPushButton("New")
         self._lb_item_source = QLabel(self._current_item_source)
 
         # Start with 'detailed' view (show optional parameters)
@@ -2096,7 +2095,6 @@ class _QtReEditor(QWidget):
         self._pb_reset = QPushButton("Reset")
         self._pb_cancel = QPushButton("Cancel")
 
-        self._pb_new_item.clicked.connect(self._pb_new_item_clicked)
         self._pb_batch_upload.clicked.connect(self._pb_batch_upload_clicked)
 
         self._pb_add_to_queue.clicked.connect(self._pb_add_to_queue_clicked)
@@ -2112,7 +2110,6 @@ class _QtReEditor(QWidget):
         hbox.addWidget(self._rb_item_instruction)
         hbox.addWidget(self._combo_item_list)
         hbox.addStretch(1)
-        hbox.addWidget(self._pb_new_item)
         hbox.addWidget(self._lb_item_source)
         vbox.addLayout(hbox)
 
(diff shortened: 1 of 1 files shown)


danielballan (approved):
Nice improvements. Good to see the batch implementation finished off.

AbbyGi (approved):

Issue #139, opened 2021-05-20
RE Widgets: turn on plan editing when item is double-clicked in the queue widget


dmgav:
Extend the functionality of the Queue Widget so that when a plan is double-clicked in the widget, it is automatically opened in the plan editor with editing enabled. 

## Expected Behavior
Double-click on a plan in Queue Widget is expected to open the plan in the plan editor.

## Current Behavior
Currently, in order to open a plan for editing the plan must be first selected in the queue. This opens the plan in the plan viewer. Clicking the `Edit` button in the plan viewer opens the plan for editing. Executing this sequence of actions for each plan may not be convenient of extensive editing is performed. An option to open plans for editing by double-clicking on the plan is intuitive wnd will improve user experience. The existing options should be preserved as well.

## Possible Solution
The main challenge with implementation of the feature is flexibility of the architecture of the widgets. Existing of the global model allows to place multiple instances of the syncronized widgets (e.g. a Queue Widget) on multiple tabs/windows. It is very likely that some of the windows/tabs (e.g. the tab that contains queue execution control widgets) will have a Queue Widget, but not the plan editor. The proposed feature should work only in the tabs/windows that contain both Queue Widget and the editor. Simple means to register editors (including custom editors) with the Queue Widget should be provided. The Queue Widget should open the plan for editing only in registered editors. If no editors are registered, then double-click events should be ignored.

dmgav:
Addressed in PR https://github.com/bluesky/bluesky-widgets/pull/145

PR #145, opened 2021-06-08
Open item in the Plan Editor widget by double-clicking on the plan

dmgav:
Improvement of user productivity by implementing easier way for opening items in the editor. Originally, the only way to open existing queue item for editing was to select item in the Plan Queue Widget and then click 'Edit' button in the Plan Viewer
tab of the editor. Changes in this PR allow to open an item in the Plan Editor by double-clicking the item in the Plan Queue Widget. This may improve productivity when small adjustments of parameters need to be done for many plans in a row (e.g. if workflow includes duplication of a plan multiple times and then setting plan parameters).

If an item is open for editing, then double-clicking on a different plan will cancel editing of currently opened plan, discard changes and open the other plan for editing. Since plan editing is expected to be trivial, it makes sense to allow easy switching between plans by reducing the number of clicks as opposed to protecting results of editing by preventing automatic cancelling of the plans that are being edited. The configurable option that allows to select if changes need to be protected may be added in the future.

## Description

In order to enable functionality, the editors must be registered with the Plan Queue Widget (see ``QtRePlanQueue.register_item_editor()``). If multiple editors are used, they need to be registered in the order starting from custom editors proceeding to general-purpose editors that will accept all plans rejected by custom editors. Once the item is double-clicked, the item is opened in the first editor that accepts it. If no editors are registered, then double-click event is ignored.

The feature is implemented purely in Qt without involving the underlying model. If multiple instances of Plan Queue widgets are used in the application, each of the instances may be calling different set of registered editors (typically editors would be located in the same window as the Plan Queue widget and some windows may not have editors).

Addresses issue https://github.com/bluesky/bluesky-widgets/issues/139

## Motivation and Context
The feature was requested by @DanOlds

### DIFF
--- a/bluesky_widgets/models/run_engine_client.py
+++ b/bluesky_widgets/models/run_engine_client.py
@@ -621,16 +621,17 @@ class RunEngineClient:
                 return copy.deepcopy(self._plan_queue_items[sel_item_pos])
         return None
 
-    def _queue_item_move(self, *, sel_items, target_item, position):
+    def _queue_items_move(self, *, sel_items, ref_item, position):
         """
-        Move the selected items above or below the target item.
+        Move the batch of selected items above or below the reference item. The target item
+        should not belong to the batch of items.
 
         Parameters
         ----------
         sel_items : list
             the list of selected item UIDs
-        target_item : str
-            UID of the target imte
+        ref_item : str
+            UID of the reference item
         position : str
             "before" - the items are moved above the target item, "after" - below the traget item
         """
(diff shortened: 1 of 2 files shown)


danielballan:
All seems sensible and clearly well thought through.

Just one part gives me pause: I prefer to avoid order-dependent registration if possible. Could we expose a list `item_editors` that, similar to `sys.path`, would allow items to be added at any position? Are there any drawbacks to making the “registry” fully public in this fashion?

dmgav:
I don't think there is a serious issue with making the registry public. Yes, I can expose the list of editors. This will allow to register editors dynamically.

danielballan (approved):

AbbyGi (approved):
Looks good. Can be merged after rebase.

PR #146, opened 2021-07-12
Move code for formatting plan parameter descriptions to `bluesky-queueserver` package

dmgav:
The code for formatting plan parameters used by plan viewer/editor widgets was moved to `bluesky-queueserver` package. The formatting code was removed from `bluesky-widgets` package. The widgets are now using function `format_text_descriptions` imported from `bluesky-queueserver`. The change is not expected to influence the behavior of the widgets. The imported function was modified in a separate PR in `bluesky-queueserver` to provide better and more consistent formatting of descriptions.

### DIFF
--- a/bluesky_widgets/models/run_engine_client.py
+++ b/bluesky_widgets/models/run_engine_client.py
@@ -6,8 +6,7 @@ import time
 import importlib
 
 from bluesky_live.event import EmitterGroup, Event
-from bluesky_queueserver.manager.comms import ZMQCommSendThreads, CommTimeoutError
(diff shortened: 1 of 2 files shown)


gwbischof (commented):
Don't we need to add a bluesky-queueserver dependency?

dmgav:
I don't think we should explicitly add `bluesky-queueserver` as a dependency before it is properly packaged and kept up-to-date.

gwbischof:
Maybe in requirements-dev?

dmgav:
`bluesky-queueserver` must be installed from source, so having it in requirements doesn't help.

PR #153, opened 2021-09-07
Line painter fix using debouncing

AbbyGi:
## Description

This PR uses the `debounce` decorator from `bluesky-live` to space out the callbacks coming from the `new_data` event so that we don't lock up the application when new data is coming in faster than the app can handle. The kafka and zmq dispatchers have been updated to process new data when it comes in and the `LOADING_LATENCY` values have been increased.

## Motivation and Context

These changes are to help ease the line painter issue where the app receiving live data would lock up because it couldn't handle the new data fast enough.

## How Has This Been Tested?

Tested interactively with replayed "live" data from the IOS beamline


AbbyGi:
Relies on https://github.com/bluesky/bluesky-live/pull/19

mrakitin (commented):
I don't have much to add, but I can't provide a very informed review as I don't know the codebase very well. Maybe worth having a meeting to go over the architecture and the structure of the code and the goals.

By the way, are the CI failures expected?

AbbyGi:
@mrakitin, the more important PR to look at would be the `bluesky-live` PR [here](https://github.com/bluesky/bluesky-live/pull/19) if you get a chance. This PR relies on changes from the `bluesky-live` PR, so CI failures are expected right now.

PR #157, opened 2021-10-19
Keep AutoLines._lines_instances dict up to date with figures/plot_builders lists

AbbyGi:
## Description
`AutoLines` now has a function (`on_figure_removed`) that updates the `_lines_instances` dict when a figure is removed or a tab is closed.

## Motivation and Context
The `AutoLines._line_instances` dict is not updated after closing a figure, so no new figures are created after closing out all the figures and adding another run.

## How Has This Been Tested?
Tested interactively by using `AutoLines` to plot a scan, closing the tab, and rerunning the scan.

### DIFF
--- a/bluesky_widgets/models/auto_plot_builders/_lines.py
+++ b/bluesky_widgets/models/auto_plot_builders/_lines.py
@@ -26,6 +26,7 @@ class AutoLines(AutoPlotter):
         # Map (stream_name, x, tuple_of_tuple_of_ys) to line of Lines instances for each group of y.
         self._lines_instances = {}
         self._max_runs = max_runs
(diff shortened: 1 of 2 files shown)


danielballan (approved):

danielballan:
One fix needed to make `flake8` happy:

https://github.com/bluesky/bluesky-widgets/runs/3944835297?check_suite_focus=true#step:5:7

Once that is addressed is CI passes* self-merge at will @AbbyGi.

*It looks like the Python 3.6 tests are failing on `main` consistently, something to do with the Jupyter view. IMO the best solution is to drop 3.6 support.

PR #159, opened 2021-11-01
Plot time relative to the start of the run

### LINKED ISSUE #156: Add "epoch" feature from LivePlot
## Expected Behavior

With `LivePlot`, when a user plots vs `"time"` (interpreted as `event['time']`) we show time relative to the `run_start_doc['time']` by default. This lets repeated scans lay nicely over each other, and it makes that scale `0` mean something useful, instead of "number of seconds since 1970". This is typically what users want, but not _always_, so we provide an option, `epoch`, which can be set to `"run"` (default) or `"unix"` (as in "The UNIX Epoch" --- seconds since 1970).

Here is the option:

https://github.com/bluesky/bluesky/blob/759f9c55dce97dc47513cca749a69dd861bdf58d/bluesky/callbacks/mpl_plotting.py#L111-L113

And where we capture the Run Start time:
https://github.com/bluesky/bluesky/blob/759f9c55dce97dc47513cca749a69dd861bdf58d/bluesky/callbacks/mpl_plotting.py#L174-L175

And where we subtract it off:

https://github.com/bluesky/bluesky/blob/759f9c55dce97dc47513cca749a69dd861bdf58d/bluesky/callbacks/mpl_plotting.py#L212-L215

## Current Behavior

The `Lines` plot builder doesn't support this yet.

## Possible Solution

In here: https://github.com/bluesky/bluesky-widgets/blob/23549f23bc615a4502fbee4b22f552af9d403442/bluesky_widgets/models/utils.py#L99-L103

1. Get `run.metadata['start']['time']` in a variable.
2. Subtract off of `namespace['time']` and, inside the loop over `reversed(stream_names)` subtract it off of every namespace[stream_name]['time']`.
3. Perhaps for now, _don't_ add an `epoch` option until/unless someone requests that feature. AFAIK the `epoch="unix"` option is never used. If we need it, we can add a parameter to `call_or_eval`, `construct_namespace` and `Lines` to pass this through and disable the subtracting off code, but that's more complexity we _may_ not need.

AbbyGi:
## Description

The run start time is subtracted from all times in the run so that anything plotted vs `time` shows the time relative to the start of the run instead of seconds since 1970 (the UNIX epoch).

## Motivation and Context

Closes #156.

## How Has This Been Tested?
Tested interactively by plotting the results of `bp.count` on simulated detectors.

## Screenshots:
time_plot

### DIFF
--- a/bluesky_widgets/models/utils.py
+++ b/bluesky_widgets/models/utils.py
@@ -94,13 +94,17 @@ def construct_namespace(run, stream_names):
     """
     namespace = dict(_base_namespace)  # shallow copy
     with lock_if_live(run):
+        run_start_time = run.metadata["start"]["time"]
         # Add columns from streams in stream_names. Earlier entries will get
         # precedence.
         for stream_name in reversed(stream_names):
             ds = run[stream_name].to_dask()
             namespace.update({column: ds[column] for column in ds})
             namespace.update({column: ds[column] for column in ds.coords})
+        namespace["time"] = namespace["time"] - run_start_time
         namespace.update({stream_name: run[stream_name].to_dask() for stream_name in stream_names})
+        for stream_name in stream_names:
+            namespace[stream_name]["time"] = namespace[stream_name]["time"] - run_start_time
     namespace.update({"run": run})
     return namespace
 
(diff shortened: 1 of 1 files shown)


danielballan:
This test failure looks related to the PR:

https://github.com/bluesky/bluesky-widgets/runs/4081964781?check_suite_focus=true#step:9:12881

Perhaps all we need to do is put `namespace["time"] = ...` inside a conditional that checks `if "time" in namespace`.

PR #163, opened 2021-11-06
FIX: Qt event loop does not execute the app in MS Windows

### LINKED ISSUE #162: Qt event loop does not execute the application in MS Windows
## Expected Behavior
The event loop created by `gui_qt()` should execute the QApplication if it is not already running.

## Current Behavior
When running the code as a python module (`python -m module_name`), it seems to work fine in my Linux environment, but in Windows the program just finishes without launching the application window.

The problem lies here:
https://github.com/bluesky/bluesky-widgets/blob/193db00e47d8ddf40a5eb68073a49545771ca8e4/bluesky_widgets/qt/_event_loop.py#L48

- In Linux `app.applicationName()` contains the name passed to the QApplication constructor `app = QApplication([app_name])`. This is what the code above is expecting.
- In Windows `app.applicationName()` contains the string `python`, therefore the if statement always evaluates to false.

On a side note, the code of the `gui_qt()` function seems to mask command line arguments instead of passing them on to the application.

## Possible Solution

a. Manually set the application name with [setApplicationName()](https://doc.qt.io/qt-5/qcoreapplication.html#applicationName-prop). I will push a PR for this.
b. Alternatively, refactor the `gui_qt` function to receive the same parameters as [QApplication](https://doc.qt.io/qt-5/qapplication.html#QApplication). That is, receive `argv` and pass it on to QApplication, instead of only receiving a single string with the application_name. I believe this would also allow to pass command line arguments to the application, which is probably not the case at the moment.

## Steps to Reproduce (for bugs)

1. `python -m bluesky_widgets.examples.qt_run_tree_view`
2. The program runs until completion without launching the application window.

## Your Environment

Windows 10, python 3.9.7, conda 4.10.3, PyQt5 5.15.4

carez:
Fixes #162

### DIFF
--- a/bluesky_widgets/qt/_event_loop.py
+++ b/bluesky_widgets/qt/_event_loop.py
@@ -36,6 +36,7 @@ def gui_qt(app_name):
         # if this is the first time the Qt app is being instantiated, we set
         # the name, so that we know whether to raise_ in Window.show()
         app = QApplication([app_name])
+        app.setApplicationName(app_name)
         global _our_app_name
         _our_app_name = app_name
     else:


carez:
The tests are failing for python 3.6 but it may well not be a problem with the PR because the same test [seems to be failing in master](https://github.com/bluesky/bluesky-widgets/runs/4085248417?check_suite_focus=true) and there's also an open PR to remove python 3.6 from tests: #161

tacaswell:
powercycled to run the tests.

PR #187, opened 2023-05-18
Enable 'pause' buttons whenever Run Engine is in 'running' state.

dmgav:
Small change in the behavior of RE Manager control widgets: 'Pause Deferred' and 'Pause Immediate' buttons are now activated whenever the Run Engine is in 'running' state. The buttons can be used to pause a running plan, started in the worker directly using Jupyter Console (worker in IPython kernel mode), bypassing the manager. 

The buttons are also activated when the manager is switched to the 'executing_queue' mode, which happens immediated after the queue is started by the manager. This makes GUI look more responsive.

## How Has This Been Tested?

The feature was manually tested.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -8,5 +8,6 @@ exclude =
     bluesky_widgets/_version.py,
     docs/source/conf.py
     bluesky_widgets/examples/pyFAI_dialog.py,
-ignore = E203, W503  # There are some errors produced by 'black', therefore unavoidable
+# There are some errors produced by 'black', therefore unavoidable    
+ignore = E203, W503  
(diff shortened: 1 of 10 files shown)


tacaswell (approved):

dmgav:
@tacaswell But I don't see the unit tests GH action to be running at all. It could be the same issue though.

tacaswell:
oh, I completely missed the tests did not run (saw green check marks but did not read what they where)

tacaswell:
There was a cron job on the tests and we made no changes to the repo for long enough the Europe got disabled.

I turned it back on

PR #188, opened 2023-06-09
Display Queue AUTOSTART mode in status monitor widget

dmgav:
Replace the indicator of the status of the queue (whether the queue is running) with the indicator of Queue AUTOSTART mode. The status of the queue is clearly visible based on the state of other widgets and the indicated manager state.

## How Has This Been Tested?

image


maffettone (approved):
LGTM. Will help clarity for BLS. Assume the test failures are pre-existing.

PR #192, opened 2024-01-30
Fixed bug in QtRemoteDispatcher

cjtitus:
## Description

The QtRemoteDispatcher class had an outdated name for one of the objects inside the dispatcher it was creating, which caused the entire class to crash when used. Fixing the name (`_bluesky_consumer.consumer` -> `_bluesky_consumer._consumer`) restores functionality.

See the code in https://github.com/bluesky/bluesky-kafka/blob/main/bluesky_kafka/__init__.py#L149 and https://github.com/bluesky/bluesky-kafka/commit/e9001b7485d12dd3f139744a3603b8a081d0ed0f which changed from `consumer` to `_consumer`

## How Has This Been Tested?
Tested with a containerized beamline running Kafka. Confirmed that QtRemoteDispatcher now instantiates correctly, and dispatches correctly.

### DIFF
--- a/bluesky_widgets/qt/kafka_dispatcher.py
+++ b/bluesky_widgets/qt/kafka_dispatcher.py
@@ -92,7 +92,7 @@ class QtRemoteDispatcher(QObject):
         while continue_polling():
             try:
                 # there should maybe be a poll method on the dispatcher
-                msg = self._dispatcher._bluesky_consumer.consumer.poll(
+                msg = self._dispatcher._bluesky_consumer._consumer.poll(
                     self._dispatcher._bluesky_consumer.polling_duration
                 )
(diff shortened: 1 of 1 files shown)


tacaswell:
95% the linting is because we did not pin the version of black and the new version changes the formatting a bit.

mrakitin (commented):
Looks good to me.
@cjtitus, do you mind applying the `black` formatting with the new version? Then we'll merge.

cjtitus:
What version of `black` do you want me to use? The actual latest one that's causing the lint check to fail? Cause then this PR goes from touching one name in one file, to touching a quarter of the code-base...

mrakitin (approved):

PR #199, opened 2024-11-19
ENH: tiled auth widget

maffettone:
A widget that uses tiled's auth mechanism to pass a user and password for read only permissions. 

## Description

The widget does a couple of things: 
1 . Creates a tiled `Context` and uses `Context.authenticate` to get off the ground.
2. Generates a temporary API key (default of 12 hours) that has read only scopes. @jmaruland: is there a risk of api key bloat, or does the server know to remove invalid API keys after a period of time?
3. Spawns a tiled client with that API key. This client will loose access after the timeout and a new login with be required.
4. A singleton state manager is holds onto the context, partial api_key, and tiled client object. 
5. Enables logging out, so that the api key is removed, the client object deleted, and the context logout. 
6. A big fat warning up top.

This should provide 2 forms of security: the ability to logout when finished, and the ability to set a timeout for the duration of your work should you forget to logout.

## Motivation and Context

Data security. Folks want the operator accounts to be able to persistently run a QT GUI that can read back data from rest. This allows for one user to login for their beam time, and not have their data accessible outside of their experiment. 

## How Has This Been Tested?

Local env on a Mac, using the example in the PR. Screenshots provided.

Screenshot 2024-11-19 at 9 25 34 AM
Screenshot 2024-11-19 at 9 52 43 AM
Screenshot 2024-11-19 at 9 54 21 AM


whs92:
@padraic-shafer are we using this with the pymca tiled integration?

PR #209, opened 2026-08-06
Add Widget that handles generation of progress bars for streamed watcher updates from the RE waiting hook

jwlodek:
Requires main branch of queueserver that includes the streaming of watcher updates feature.


jwlodek:
Example of how this looks in practice:

image

